PR #7, opened 2021-11-29
Support Ruby 3.0 kwarg-handling within `.delay` API

### LINKED ISSUE #6: Ruby 3 support
Ruby 3 changed the way it handles keyword arguments, and is a [roadblock with delayed_job](https://github.com/collectiveidea/delayed_job/pull/1130). I checked the relevant bits in the source of this gem (at least the bits that used to be relevant in delayed_job) and didn't see any changes or fixes for Ruby 3 keyword argument handling.

https://github.com/Betterment/delayed/blob/2d296b9f49c24c99b91d8bf78690a20436811bae/lib/delayed/message_sending.rb#L11-L13
https://github.com/Betterment/delayed/blob/2d296b9f49c24c99b91d8bf78690a20436811bae/lib/delayed/performable_method.rb#L25-L27

While the gemspec suggests Ruby 3 should work, does it really?

smudge:
/domain @effron @samandmoore 
/platform @effron @samandmoore 

Resolves #6. While we already fully support Ruby 3.0 kwargs via ActiveJob, the legacy `delay` and `handle_asynchronously` APIs did not support the new [separation of positional and keyword arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) introduced in 2.7 and enforced in 3.0.

This means that methods accepting both positional and keyword arguments (e.g. `def foo(a, b:)`) would fail with a `wrong number of arguments (given 2, expected 1; required keyword:)` error on Ruby 3. In order to resolve this, this PR changes `Delayed::PerformableMethod` to handle kwargs separately from args, with a backwards compatibility check for any Ruby 2.6 methods that do not accept keyword arguments. It should also support jobs that were enqueued by the prior `delayed` gem version (where the new `kwargs` accessor would be nil, and `args` contains the kwargs as its last item).

### DIFF
--- a/lib/delayed/message_sending.rb
+++ b/lib/delayed/message_sending.rb
@@ -18,16 +18,6 @@ module Delayed
       DelayProxy.new(PerformableMethod, self, options)
     end
     alias __delay__ delay
-
-    def send_later(method, *args)
-      warn '[DEPRECATION] `object.send_later(:method)` is deprecated. Use `object.delay.method'
-      __delay__.__send__(method, *args)
-    end
-
-    def send_at(time, method, *args)
-      warn '[DEPRECATION] `object.send_at(time, :method)` is deprecated. Use `object.delay(:run_at => time).method'
-      __delay__(run_at: time).__send__(method, *args)
-    end
   end
 
   module MessageSendingClassMethods
--- a/lib/delayed/message_sending.rb
+++ b/lib/delayed/message_sending.rb
@@ -8,8 +8,8 @@ module Delayed
       @options = options
     end
 
-    def method_missing(method, *args)
-      Job.enqueue({ payload_object: @payload_class.new(@target, method.to_sym, args) }.merge(@options))
+    def method_missing(method, *args, **kwargs)
+      Job.enqueue({ payload_object: @payload_class.new(@target, method.to_sym, args, kwargs) }.merge(@options))
     end
   end
 
@@ -36,7 +36,7 @@ module Delayed
       punctuation = $1 # rubocop:disable Style/PerlBackrefs
       with_method = "#{aliased_method}_with_delay#{punctuation}"
       without_method = "#{aliased_method}_without_delay#{punctuation}"
-      define_method(with_method) do |*args|
+      define_method(with_method) do |*args, **kwargs|
         curr_opts = opts.clone
         curr_opts.each_key do |key|
           next unless (val = curr_opts[key]).is_a?(Proc)
@@ -47,7 +47,7 @@ module Delayed
                              val.call
                            end
         end
-        delay(curr_opts).__send__(without_method, *args)
+        delay(curr_opts).__send__(without_method, *args, **kwargs)
       end
 
       alias_method without_method, method
--- a/lib/delayed/performable_method.rb
+++ b/lib/delayed/performable_method.rb
@@ -1,8 +1,8 @@
 module Delayed
   class PerformableMethod
-    attr_accessor :object, :method_name, :args
+    attr_accessor :object, :method_name, :args, :kwargs
 
-    def initialize(object, method_name, args)
+    def initialize(object, method_name, args, kwargs)
       raise NoMethodError, "undefined method `#{method_name}' for #{object.inspect}" unless object.respond_to?(method_name, true)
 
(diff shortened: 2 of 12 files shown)


lib/delayed/message_sending.rb
```diff
@@ -18,16 +18,6 @@ def delay(options = {})
...
-
-    def send_later(method, *args)
-      warn '[DEPRECATION] `object.send_later(:method)` is deprecated. Use `object.delay.method'
-      __delay__.__send__(method, *args)
-    end
-
-    def send_at(time, method, *args)
-      warn '[DEPRECATION] `object.send_at(time, :method)` is deprecated. Use `object.delay(:run_at => time).method'
-      __delay__(run_at: time).__send__(method, *args)
-    end
```
smudge, line 30: Removing these -- they were already deprecated prior to our fork.

effron:
<<domainLGTM

spec/performable_mailer_spec.rb
```diff
@@ -1,68 +1,78 @@
...
-          MyMailer.with(foo: 1, bar: 2).signup('john@example.com').delay
+          MyMailer.signup('john@example.com').delay
         }.to raise_error(RuntimeError)
       end
     end
   end
+
+  if defined?(ActionMailer::Parameterized::Mailer)
+    describe ActionMailer::Parameterized::Mailer do
+      describe 'delay' do
```
smudge, line 54: I didn't actually add this section -- I just reorganized the file a little.

samandmoore:
<<platformlgtm

nanda-prbot (approved):
Approved! :dancers: :heart_eyes_cat: :ok_woman:

lib/delayed/performable_method.rb
```diff
@@ -23,7 +24,13 @@ def display_name
     end
 
     def perform
-      object.send(method_name, *args) if object
+      return unless object
+
+      if kwargs.nil? || (RUBY_VERSION < '2.7' && kwargs.empty?)
```
smudge, line 29: There's [a proposed PR](https://github.com/collectiveidea/delayed_job/pull/1130) in the upstream repo, but for this PR I took a very different strategy, because we want to both maintain Ruby 2.6 support and support the Ruby 3 case where both positional and keyword arguments are present. Holding onto `kwargs` separately resolves edge cases around argument auto-merging behaviors, but then on Ruby 2.6 we need to fall back in the `**{}` case (since on 2.6 and below that causes `{}` to get passed as a positional argument).

smudge: But I think that this should resolve https://github.com/collectiveidea/delayed_job/issues/1134 and might be worth upstreaming, if someone gets a chance.

samandmoore:
nice. glad you tested it :)

domainltm

samandmoore:
ugh domainlgtm 

curse this keyboard!

nanda-prbot (approved):
Approved! :pray: :star: :ok_woman:

smudge:
Sorry for the review churn! I've been testing this in the context of our largest Rails monorepo, and seem to have ironed out the kinks, so I'd like to cut a new version all in one go.

samandmoore:
domainlgtm

nanda-prbot (approved):
Approved! :clap: :pizza: :boom:

lib/delayed/performable_method.rb
valerauko, line 29: You can check if the callee uses kwargs. First get the method object with `object.method(method_name)` and check its parameter list for `:key`, `:keyreq` or `:keyrest`. (From experience, Rails Mailer magical methods end up with a single `:rest` argument that you'd need to account for as well.)

effron:
domainLGTM

lib/delayed/performable_method.rb
smudge: In this case, I don't think we want it to actually care about the arity of the method it's about to call -- only the arity that was provided to it by the enqueuing call. (For example, if I do `obj.delay.something(test: 123)` on a method that _does not accept kwargs_, I want to ensure that the job raises an error and tells me that my arguments are wrong.)

So the fix here is really just to prevent `**{}` from breaking on Ruby 2.6 (because trying to pass that as a kwarg results in it passing `{}` as a positional arg). This quirk was fixed in 2.7.

PR #14, opened 2022-02-11
Fix main builds: appease the linter

smudge:
/domain @samandmoore @effron 
/no-platform

There are a couple linter rules complaining. This will ensure that anyone who branches off of main won't have to fix them independently.

This also fixes the Ruby 3.0 build so that it doesn't inadvertently run with Ruby 3.1 in CI.


jmileham:
<< domain lgtm

nanda-prbot (approved):
Approved! :avocado: :ok_woman: :sparkler:

Issue #23, opened 2023-01-20
Difference in error behavior when a job is undefined


daichi5:
Hi! We have migrated `DelayedJob` to `Delayed` and are running it.
While observing the worker, we came across an error and found a difference in behavior.

Here is the stack trace in the `last_error` column.

<details>

> Job failed to load: undefined class/module Sample. Handler: \"--- !ruby/object:Delayed::JobWrapper\
> job_data:\
>   job_class: Sample\
>   job_id: 0f9f613a-682b-439c-8169-d45a4984e8eb\
>   provider_job_id: \
>   queue_name: v2\
>   priority: \
>   arguments: &2 []\
>   executions: 0\
>   exception_executions: &1 {}\
>   locale: en\
>   timezone: UTC\
>   enqueued_at: '2023-01-20T11:51:44Z'\
> job: !ruby/object:Sample\
>   arguments: []\
>   job_id: 0f9f613a-682b-439c-8169-d45a4984e8eb\
>   queue_name: v2\
>   priority: \
>   executions: 0\
>   exception_executions: *1\
>   timezone: UTC\
>   provider_job_id: \
>   serialized_arguments: *2\
>   locale: en\
>   enqueued_at: '2023-01-20T11:51:44Z'\
> \"
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/backend/base.rb:84:in `rescue in payload_object'
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/backend/base.rb:81:in `payload_object'
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/backend/base.rb:133:in `max_run_time'
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/worker.rb:194:in `max_run_time'
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/worker.rb:137:in `block in run'
> /usr/local/lib/ruby/3.0.0/benchmark.rb:308:in `realtime'
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/worker.rb:136:in `run'
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/worker.rb:210:in `block in run_job'
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/lifecycle.rb:61:in `block in initialize'
> /usr/local/bundle/gems/delayed-0.4.0/lib/delayed/lifecycle.rb:66:in `execute'
[...]

PR #24, opened 2023-01-20
Reduce handler size in some ActiveJob cases, and fix deserialization inconsistency

smudge:
This should address #23.

The issue here is that some ActiveJob-enqueued jobs would include the `@job` ivar in their serialized handler, which is redundant with `@job_data` and only existed for memoization purposes. We can use Ruby's `encode_with` API to ensure that we only encode `@job_data` into the handler.

Including `@job` also had the side-effect of causing `DeserializationErrors` and permanently failing jobs in cases where there would otherwise be a retryable `NameError`. Whether a missing constant _should_ result in a `DeserializationError` is kind of a separate question, but since `ActiveJob`'s stance seems to be to encode the class name as a string and only `.constantize` it inside of the perform, I think that it's not surprising behavior to let the `NameError` (and subsequent retries) occur. `DeserializationError` is really supposed to be only for cases where we fundamentally can't even attempt the job and allow the normal pattern of exception-handling to occur.


effron (approved):
Changes look good to me! Only suggestion would be to add a test that explicitly makes sure we raise a `NameError` and not a `DeserializationError` in the circumstances described here https://github.com/Betterment/delayed/issues/23

effron (dismissed):
domainLGTM

PR #28, opened 2023-10-06
Rails 7.1, Ruby 3.1, and Ruby 3.2

smudge:
This adds Rails 7.1, Ruby 3.1, and Ruby 3.2 to the build matrix.

The gem already works with these versions, but now they are formally tested. (As such, I bumped the gem by a patch version, to signify that this is the first verified-compatible release.)


.github/workflows/ci.yml
```diff
@@ -8,15 +8,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.6', '2.7', '3.0']
+        ruby: ['2.6', '2.7', '3.0', '3.1', '3.2']
```
smudge, line 11: I'd like to drop Ruby < 3 support in the next major release.

spec/delayed/serialization/active_record_spec.rb
```diff
@@ -3,13 +3,13 @@
...
     expect {
-      YAML.load(Story.create.to_yaml)
+      YAML.load_dj(Story.create.to_yaml)
     }.not_to raise_error
   end
 
   it "loads classes even if not in default scope" do
     expect {
-      YAML.load(Story.create(scoped: false).to_yaml)
+      YAML.load_dj(Story.create(scoped: false).to_yaml)
```
smudge, line 12: This spec just exists to make sure that ActiveRecords work with the psych ext. The new safe YAML.load in Ruby 3.1 and 3.2 requires extra configuration if you want to (de)serialize custom classes, but instead we can use the patched method that this gem actually uses for the same effect as before.

spec/delayed/active_job_adapter_spec.rb
```diff
@@ -34,15 +34,16 @@ def perform; end
...
+        /  priority: ?\n/,
         "  arguments: []\n",
         "  executions: 0\n",
         ("  exception_executions: {}\n" if ActiveJob::VERSION::MAJOR >= 6),
         "  locale: en\n",
-        ("  timezone: \n" if ActiveJob::VERSION::MAJOR >= 6),
-        ("  enqueued_at: '2023-01-20T18:52:29Z'\n" if ActiveJob::VERSION::MAJOR >= 6),
+        (/  timezone: ?\n/ if ActiveJob::VERSION::MAJOR >= 6),
+        (/  enqueued_at: '2023-01-20T18:52:29(\.\d+)?Z'\n/ if ActiveJob::VERSION::MAJOR >= 6),
+        (/  scheduled_at: ?\n/ if ActiveJob::VERSION::MAJOR >= 7 && ActiveJob::VERSION::MINOR >= 1),
```
smudge, line 46: This is all some minor formatting variances introduced in newer Rubies & Rails versions.

effron (approved):
domainLGTM

PR #33, opened 2023-12-28
Drop EOL Ruby and Rails versions support

Irving-Betterment:
**Summary of changes:**

- Drop end of life [Ruby](https://endoflife.date/ruby) and [Rails](https://endoflife.date/rails) versions support.
- Address rubocop offenses (detailed per commit).

/task https://app.asana.com/0/1206095065335149/1206250886010422/f


lib/delayed/performable_method.rb
```diff
@@ -24,13 +24,7 @@ def display_name
...
 
     def perform
-      return unless object
-
-      if kwargs.nil? || (RUBY_VERSION < '2.7' && kwargs.empty?)
-        object.send(method_name, *args)
-      else
-        object.send(method_name, *args, **kwargs)
-      end
+      object.send(method_name, *args, **kwargs) if object
```
Irving-Betterment, line 27: [I reverted this guard](https://github.com/Betterment/delayed/pull/7/files) after Ruby <3.0 support is not longer required.

lib/delayed/performable_method.rb
```diff
@@ -24,13 +24,7 @@ def display_name
     end
 
     def perform
-      return unless object
-
-      if kwargs.nil? || (RUBY_VERSION < '2.7' && kwargs.empty?)
```
Irving-Betterment, line 29: [I reverted this guard](https://github.com/Betterment/delayed/pull/7/files) after Ruby <3.0 support is not longer required.

effron: do we still need the `kwargs.nil?` check?

Irving-Betterment: We do, I committed back the support of Ruby 3 positional and keyword arguments.

smudge:
@Irving-Betterment For this gem, I think we'd like to keep supporting older versions for the time being. It's intended as a drop-in replacement for `delayed_job`, and keeping more relaxed version constraints reduces the amount of work it would take to migrate over. I think we should consider dropping version support only if maintaining the older version support becomes a particular pain point.

Irving-Betterment:
Closed as we want to keep supporting older versions for now.

PR #34, opened 2024-01-08
Address dependabot alerts

smudge:
Committing the lockfile triggered a bunch of dependabot errors, and I realized that the versions committed were very out of date, so this updates the lockfile with newer versions.

/no-platform

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -206,4 +206,4 @@ DEPENDENCIES
   zeitwerk
 
 BUNDLED WITH
-   2.4.11
+   2.4.13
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -126,4 +206,4 @@ DEPENDENCIES
   zeitwerk
 
 BUNDLED WITH
-   2.4.13
+   2.4.11
```
effron, line 209: should we avoid decreasing the bundler version?

effron (dismissed):
just one nonblocking comment

domainLGTM!

Irving-Betterment (approved):
domainlgtm! Other than the bundler version Harry pointed out.

effron (approved):
domainLGTM

PR #38, opened 2024-01-31
Fix the lockfile / linter build step

smudge:
/no-platform

We commit the lockfiles now, so version changes need a re-bundle. (I missed this in the last PR.) Once this is in I can cut a new release.


jmileham (approved):
domain lgtm

PR #39, opened 2024-04-03
Allow configuring default priority value at midpoint of range

tkoar:
### Summary
This PR Adds a configuration value (`.assign_at_midpoint`) to `Delayed::Priority` 

The intent is to allow users to change the default priority of a job (i.e. when using a named priority) to be the middle of the configured priority ranges -- derived from the default or custom priority names -- as opposed to the starting value of the range.

By allowing a default value that is in the middle of the range, users will be able to bump the priority of some jobs such that they remain in their assigned queue but take priority over jobs of the same named priority.

**Note**: since the last priority uses a range that is infinite, it's not possible to derive the midpoint. So, for now, I've defaulted to adding 5, since that matches our defaulted priority values.


lib/delayed/priority.rb
```diff
@@ -82,6 +82,15 @@ def alerts=(alerts)
         @alerts = alerts&.sort_by { |k, _| names.keys.index(k) }&.to_h
       end
 
+      def assign_at_midpoint?
+        @assign_at_midpoint || false
+      end
+
+      def assign_at_midpoint=(assign_at_midpoint=false)
```
smudge, line 89:
```suggestion
      def assign_at_midpoint=(assign_at_midpoint)
```

Argument defaults don't work on assignment methods.

lib/delayed/priority.rb
```diff
@@ -82,6 +82,15 @@ def alerts=(alerts)
...
+        @assign_at_midpoint = assign_at_midpoint
+        @assign_at_midpoint
```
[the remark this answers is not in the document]
tkoar: ah right, this was just missed cleanup

lib/delayed/priority.rb
tkoar: Yeah, that makes sense. You probably can't write `thing.attribute=` anyway

spec/delayed/priority_spec.rb
```diff
@@ -43,6 +45,17 @@
...
 
+    context 'when assign_at_midpoint is set to true' do
+      let(:assign_at_midpoint) { true }
+
+      it 'returns the midpoint value' do
+        expect(described_class.interactive).to eq 5
+        expect(described_class.user_visible).to eq 15
+        expect(described_class.eventual).to eq 25
+        expect(described_class.reporting).to eq 35
+      end
```
smudge, line 56: I think we also want to test the inverse of these.

There is a test down near line 110 that looks like this:
```ruby
  it 'provides the name of the priority range' do
    expect(described_class.new(0).name).to eq :interactive
    expect(described_class.new(3).name).to eq :interactive
    expect(described_class.new(10).name).to eq :user_visible
    expect(described_class.new(29).name).to eq :eventual
    expect(described_class.new(999).name).to eq :reporting
    expect(described_class.new(-123).name).to eq nil
  end
```

We should test with the midpoint feature enabled too, to ensure that the ranges still map to the same words. (E.g. 0-9 maps to interactive, 10-19 maps to user_visible, etc).

As it stands I suspect that 5-14 will map to interactive and 15-24 will map to user_visible (etc) and I want to either prove that or disprove that with the test coverage. 🙂

lib/delayed/priority.rb
```diff
@@ -82,12 +82,39 @@ def alerts=(alerts)
         @alerts = alerts&.sort_by { |k, _| names.keys.index(k) }&.to_h
       end
 
+      def assign_at_midpoint?
+        @assign_at_midpoint || false
+      end
+
+      def assign_at_midpoint=(assign_at_midpoint)
+        @assign_at_midpoint = assign_at_midpoint
+      end
```
tkoar, line 91: I can take a shot at adding some documentation for this if this is the direction we end up going in

spec/delayed/priority_spec.rb
tkoar: Good call! this was not working.. the ranges were still being built using the new mid-point values. Fixed that by creating a new mapping altogether.

lib/delayed/priority.rb
```diff
@@ -118,7 +145,7 @@ def method_missing(method_name, *args)
 
     def initialize(value)
       super()
-      value = self.class.names[value] if value.is_a?(Symbol)
+      value = self.class.names_to_priority[value] if value.is_a?(Symbol)
```
smudge, line 148: Nice, so now the underlying `names` map remain unchanged (and keeps the original ranges), but when we instantiate (e.g. `new(:interactive)`) we the `names_to_priority` map instead of the `names` map. 👍

lib/delayed/priority.rb
```diff
@@ -82,12 +82,39 @@ def alerts=(alerts)
...
+        @assign_at_midpoint = assign_at_midpoint
+      end
+
       def ranges
         @ranges ||= names.zip(names.except(names.keys.first)).each_with_object({}) do |((name, lower), (_, upper)), obj|
           obj[name] = (lower...(upper || Float::INFINITY))
         end
       end
 
+      def names_to_priority
```
tkoar, line 99: Will add some tests for this too, but needed to make it public to access it from the initialize method

lib/delayed/priority.rb
```diff
@@ -98,13 +114,20 @@ def default_alerts
         @names ? {} : DEFAULT_ALERTS
       end
 
+      def names_to_midpoint_priority
+        names.each_cons(2).to_h { |(name, priority_value), (_, next_priority_value)|
+          midpoint = priority_value.to_i + ((next_priority_value.to_i - priority_value.to_i).to_d / 2).ceil
```
tkoar, line 119: I think we could avoid the `to_i` calls by adding the math operator functions for addition & subtraction (similar to the custom comparison operator), but wanted to get thoughts on that first

README.md
```diff
@@ -491,6 +500,14 @@ Delayed::Worker.min_priority = nil
...
 ```
 
+Job priorities can specified by using the name of the desired range (i.e. :user_visible).
+By default, the value for a named priority will be the first value in that range.
+To set each priority's default value to the middle of its range (i.e. 15 for :user_visible), Delayed::Priority can be configured with:
+
+```ruby
+Delayed::Priority.assign_at_midpoint = true
+```
+
```
smudge, line 510: Yup, this makes sense. At some point we should consider making this be the default, but opt-in is good for now.

lib/delayed/priority.rb
```diff
@@ -147,6 +173,14 @@ def <=>(other)
       to_i <=> other
     end
 
+    def -(other)
+      to_i - other.to_i
+    end
+
+    def +(other)
+      to_i + other.to_i
+    end
```
smudge, line 182: I was thinking about what it meant to always call `to_i` on other, and also what it means to always respond with an integer. For example:

```
Priority.new(1) + Priority.new(10)
```

I would expect that to return `Priority.new(11)`, not `11`.

And also these examples:

```ruby
Priority.new(10) + 1
```

Should also probably return `Priority.new(11)`.

But I'm not sure what this should return:

```ruby
Priority.new(10) + 0.9
```
(Should it round to `Priority.new(11)`, or raise an error?)

Maybe we don't need to handle that last case for now.

So I think that leaves us with this:

```suggestion
    def -(other)
      new(to_i - other.to_i)
    end

    def +(other)
      new(to_i + other.to_i)
    end
```

Though we could do what we do in `<=>` and cast other to_i only if it's a priority:

```ruby
other = other.to_i if other.is_a?(self.class)
```

WDYT?

tkoar: Oh yeah that's a good call! We should return a Priority instance, since that is the caller.

Hmm yeah the math complexities are interesting. I would say we probably kick the can on dealing with floats? Right now the `initialize` method (and the mid_point method) will ensure all instances have integer values.

smudge (approved):
domain LGTM && platform LGTM

Issue #41, opened 2024-06-14
Possible unthrottled spinloop when jobs fail to deserailize


smudge:
Working with @emmaoberstein on setting up a new queue, we discovered that there is a condition when the worker will get stuck in a spinloop, running the pickup query multiple times per second (effectively disobeying the `sleep_delay` config).

The likeliest reproduction would be to enqueue a job that fails deserialization completely, or that otherwise prevents the usual job cleanup (e.g. it crashes the thread).

Here's where the `work_off` method hits the spinloop:

```ruby
    def work_off(num = 100)
      success = Concurrent::AtomicFixnum.new(0)
      failure = Concurrent::AtomicFixnum.new(0)

      num.times do
        jobs = reserve_jobs
        break if jobs.empty? # jobs is not empty

        pool = Concurrent::FixedThreadPool.new(jobs.length)
        jobs.each do |job|
          pool.post do
            # Exception encountered when `payload_object` is first called, e.g. job fails to deserialize
            # - success and failure are never incremented
            # - job remains in queue and is immediately returned by the next `reserve_jobs` without waiting
          end
        end

        pool.shutdown
        pool.wait_for_termination

        break if stop?
      end

      [success, failure].map(&:value)
    end
```

There a few ways I could think to fix this:
1. Add a new "inner loop" delay that sets a minimum amount of time in between each iteration of `num.times do`.
2. Bail from the loop if neither success nor failure were incremented (i.e. no work got done).
3. Ensure that job cleanup happens in all cases (except for complete loss of DB connection), to ensure that `reserve_jobs` won't immediately return the same job again (due to exponential backoff).

All of these feel fairly reasonable, though I'd be inclined to explore the second and third. (Adding a new delay would require more tuning & testing and would not actually address the underlying failure mode for the job.) So, actually, I'd want to start with the third option, since it would likely also address the remaining issue in #23.

smudge:
Should be resolved by #42

PR #42, opened 2024-06-26
Fix spin-loop/cleanup failure mode within run loop

### LINKED ISSUE #23: Difference in error behavior when a job is undefined

smudge:
This ensures that exceptions raised in thread callback hooks are rescued and properly mark jobs as failed.

This is also a good opportunity to change the `num` argument (of `work_off(num)`) to mean number of jobs (give or take a few due to `max_claims`), not number of iterations. Previously (before threading was introduced) I think it meant number of jobs (though jobs and iterations were 1:1). I would not have done this before the refactor, because there was no guarantee that one of `success` or `failure` would be incremented (the thread might crash for many reasons). Now, we only increment `success` and treat `total - success` as the "failure" number when we return from the method.

Fixes #23 and #41

This is also a prereq for a resolution I'm cooking up for #36


lib/delayed/worker.rb
```diff
@@ -89,22 +89,17 @@ def on_exit!
...
-      num.times do
+      while total < num
         jobs = reserve_jobs
         break if jobs.empty?
 
+        total += jobs.length
         pool = Concurrent::FixedThreadPool.new(jobs.length)
         jobs.each do |job|
           pool.post do
-            run_thread_callbacks(job) do
```
smudge, line 101: The explanation might make more sense if I put it here next to the related code.

Previously, if `run_thread_callbacks` crashed for any reason, neither `success` nor `failure` would be incremented. Furthermore, the job cleanup (down at the end of `run(job)`) would never occur, so attempts would not be incremented and run_at would not be bumped, making the job immediately available for pickup again.

This is what I mean by the "spinloop". A job gets picked up, its thread crashes, no cleanup occurs, it gets immediately picked up again, etc. Pushing the `run_thread_callbacks` call into `run_job` allows us to clean up the job if, e.g., it fails to deserialize, or if one of our callbacks fails to connect to a secondary resource (as was the case in #41).

lib/delayed/worker.rb
```diff
@@ -114,38 +109,41 @@ def work_off(num = 100)
...
+        job_say job, format('COMPLETED after %.4f seconds', run_time)
       end
-      job_say job, format('COMPLETED after %.4f seconds', run_time)
       true # did work
     rescue DeserializationError => e
       job_say job, "FAILED permanently with #{e.class.name}: #{e.message}", 'error'
 
       job.error = e
       failed(job)
+      false # work failed
```
smudge, line 146: This was a bug that improperly reported deserialization errors in the `success` number, which only impacted the logging output.

.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.6', '2.7', '3.0', '3.1', '3.2']
+        ruby: ['2.7', '3.0', '3.1', '3.2']
```
smudge, line 11: we can make this more formal, but I wanted to unblock this build for now, without the linter churn that comes from actually changing the min supported Ruby.

lib/delayed/worker.rb
```diff
@@ -89,22 +89,17 @@ def on_exit!
     # Exit early if interrupted.
     def work_off(num = 100)
       success = Concurrent::AtomicFixnum.new(0)
-      failure = Concurrent::AtomicFixnum.new(0)
+      total = 0
 
-      num.times do
+      while total < num
```
samandmoore, line 94: So I'm not sure how likely it is to matter but this change means that we will keep looping until we see num jobs which means that if the queue becomes empty before we hit num we will keep looping, right? 

Is that okay or desired?

samandmoore (commented):
Change looks good as a solve for the issue. Just one question about the other behavior change!

lib/delayed/worker.rb
samandmoore: ah! i missed that. excellent.

samandmoore (approved):
domainlgtm

samandmoore (approved):
platformlgtm

PR #44, opened 2024-08-20
update rexml + nokogiri to fix security alerts

samandmoore:
just what it says. we've got some dependabot alerts for these deps, so upgrading to fix that.


smudge (approved):
domain LGTM && platform LGTM

PR #48, opened 2024-12-18
Add a config for enforcing a minimum job reserve interval (default: 0 for now)

smudge:
This is related to #41 and #42 insofar as a kind of DB-bound "spinloop" is still possible if a worker picks up jobs that take so little time that the worker immediately turns around and asks for more. As of now, there has been no way to tune the amount of time a worker should wait in between _successful_ iterations of its run loop. 

This introduces a configuration (`min_reserve_interval`) specifying a minimum number of seconds (default: 0 as this is not a major release) that a worker should wait in between _successful_ job reserve queries. An existing config (`sleep_delay`) is still used to define the number of seconds (default: 5) that a worker should wait in between _unsuccessful_ job reserve attempts (i.e. the queue is empty).

The job execution time is subtracted from `min_reserve_interval` when the worker sleeps, and if jobs take more than `min_reserve_interval` to complete than the worker will not sleep before the next reserve query.

/no-platform

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,10 @@ Delayed::Worker.read_ahead = 5
 
 # If a worker finds no jobs, it will sleep this number of seconds in between attempts:
 Delayed::Worker.sleep_delay = 5
+
+# Until version 1.0, the worker will not sleep at all between attemps if it finds jobs.
+# This can be configured by setting the minimum reserve interval:
+Delayed::Worker.min_reserve_interval = 0.5 # seconds
 ```
 
 If a job fails, it will be rerun up to 25 times (with an exponential back-off). Jobs will also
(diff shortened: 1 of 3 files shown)


effron (approved):
domainLGTM!

jmileham:
Should a default behavior change like this be reserved for a major version? Like introduce the param at 0 seconds in 0.6 and then pop it to 1s at 1.0?

jmileham (approved):
domain lgtm too! Will leave platform to somebody closer/more current, if we want that. But I also think Nathan's currency satisfies that requirement for me for no-platform.

README.md
```diff
@@ -432,6 +432,10 @@ Delayed::Worker.read_ahead = 5
 
 # If a worker finds no jobs, it will sleep this number of seconds in between attempts:
 Delayed::Worker.sleep_delay = 5
+
+# Until version 1.0, the worker will not sleep at all between attemps if it finds jobs.
+# This can be configured by setting the minimum reserve interval:
+Delayed::Worker.min_reserve_interval = 0.5 # seconds
```
effron, line 438: why 0.5 and not 0?

effron: I see!

PR #49, opened 2025-01-14
Rails 8 support

rzane:
This PR runs our test suite against Rails 8 on CI.

There was only one major change. `ActiveSupport::ProxyObject` was deprecated in 7.2 and removed in 8.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,7 @@
 # Changelog
-All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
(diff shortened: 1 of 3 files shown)


lib/delayed/message_sending.rb
```diff
@@ -1,7 +1,8 @@
-require 'active_support/proxy_object'
-
 module Delayed
-  class DelayProxy < ActiveSupport::ProxyObject
+  class DelayProxy < BasicObject
+    undef_method :==
+    undef_method :equal?
```
rzane, line 4: https://github.com/rails/rails/blob/v7.2.2.1/activesupport/lib/active_support/proxy_object.rb

PR #53, opened 2025-04-02
[fix] Thread callback ordering

smudge:
/no-platform
/domain @argvniyx-enroute @mavenraven 

PR https://github.com/Betterment/delayed/pull/42 inadvertently flipped the order of `:thread` and `:perform`, and also pushed `:thread` far enough in that cleanup steps would happen after the `with_connection`'s `end` block in the connection plugin.

This introduces the possibility of thread safety issues, or connections being held longer than intended (exhausting the connection pool / connection limits).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,3 +20,6 @@ Rails/EnvironmentVariableAccess:
 
 RSpec/IndexedLet:
   Enabled: false
+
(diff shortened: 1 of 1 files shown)


argvniyx-enroute (approved):
domainlgtm

PR #54, opened 2025-04-03
0.8.0 - `Priority#to_f`, plus some regression coverage

smudge:
I'm adding regression coverage for the changes made in #53, and making one additional change in line with the way that the `:perform` callback [works in the `delayed_job` gem](https://github.com/collectiveidea/delayed_job/blob/ea4879dd3c2f3f5aa7f616a6f0fd58f2bbc2a481/lib/delayed/worker.rb#L313) - there is also now a test that will more explicitly check the order of callback events and behavior on job failure.

Plus, while debugging in an app, I noticed that datadog tries to cast priority to a float, which was raising a method undefined error. So I added `Delayed::Priority#to_f` and added a test to cover all the explicit conversion methods (`to_s`, `to_i`, `to_f`, `to_d`).

/no-platform

### DIFF
--- a/spec/worker_spec.rb
+++ b/spec/worker_spec.rb
@@ -242,6 +242,7 @@ describe Delayed::Worker do
 
     it 'wraps perform and cleanup, even when perform raises' do
       events = []
+      last_error = nil
 
       plugin = Class.new(Delayed::Plugin) do
         callbacks do |lifecycle|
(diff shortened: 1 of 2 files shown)


spec/worker_spec.rb
```diff
@@ -239,5 +239,35 @@
...
+          end
+        end
+      end
+
+      Delayed.plugins << plugin
+
+      Delayed::Job.enqueue ErrorJob.new
+      described_class.new.work_off
+
+      expect(events).to eq %i(thread_start perform_start perform_end thread_end)
```
argvniyx-enroute, line 269: 🎉

argvniyx-enroute (approved):
domainlgtm

PR #55, opened 2025-05-07
Delayed 1.0: Lifecycle & job cleanup improvements

smudge:
This PR bumps us to `1.0.0` because in cleaning up our lifecycle hooks, I made a breaking change to remove the `job` argument from the `:thread` lifecycle callback. Coupling threads & jobs 1:1 was an implementation decision, and we are restricting ourselves by baking that decision into the callback hooks. This also ensures that consuming apps do not have access to the job instance during `:thread` callbacks, making job-level cleanup slightly easier for the framework.

In addition to that change, I wrote more extensive tests around the full callback lifecycle under several different failure conditions to ensure that jobs get cleaned up in all cases, and adjusted the placement of `rescue` calls to match my own expectations of how this framework should handle these failures. This rework aims to prevent "spinloop" scenarios where, due to failed job cleanup, workers get stuck attempting to work off the same jobs over & over. (Even with `locked_at`, these workers can slowly degrade if enough jobs are consistently stuck in a locked state past the lock timeout.)

/no-platform

### DIFF
--- a/lib/delayed/lifecycle.rb
+++ b/lib/delayed/lifecycle.rb
@@ -8,7 +8,7 @@ module Delayed
       perform: %i(worker job),
       error: %i(worker job),
       failure: %i(worker job),
-      thread: %i(worker job),
+      thread: [:worker],
       invoke_job: [:job],
     }.freeze
 
--- a/lib/delayed/worker.rb
+++ b/lib/delayed/worker.rb
@@ -144,13 +144,10 @@ module Delayed
         job_say job, format('COMPLETED after %.4f seconds', run_time)
         true # did work
       rescue DeserializationError => e
-        job_say job, "FAILED permanently with #{e.class.name}: #{e.message}", 'error'
-
-        job.error = e
-        failed(job)
+        handle_unrecoverable_error(job, e)
         false # work failed
       rescue Exception => e # rubocop:disable Lint/RescueException
-        self.class.lifecycle.run_callbacks(:error, self, job) { handle_failed_job(job, e) }
+        handle_erroring_job(job, e)
         false # work failed
       end
     end
@@ -204,10 +201,21 @@ module Delayed
       " (queue=#{queue})" if queue
     end
 
-    def handle_failed_job(job, error)
+    def handle_erroring_job(job, error)
+      self.class.lifecycle.run_callbacks(:error, self, job) do
+        job.error = error
+        job_say job, "FAILED (#{job.attempts} prior attempts) with #{error.class.name}: #{error.message}", 'error'
+        reschedule(job)
+      rescue Exception => e # rubocop:disable Lint/RescueException
+        handle_unrecoverable_error(job, e)
+      end
+    end
+
+    def handle_unrecoverable_error(job, error)
+      job_say job, "FAILED permanently with #{error.class.name}: #{error.message}", 'error'
+
       job.error = error
-      job_say job, "FAILED (#{job.attempts} prior attempts) with #{error.class.name}: #{error.message}", 'error'
-      reschedule(job)
+      failed(job)
     end
 
     # The backend adapter may return either a list or a single job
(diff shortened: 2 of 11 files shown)


spec/helper.rb
```diff
@@ -126,10 +133,7 @@ class SingletonClass
     read_ahead_was = Delayed::Worker.read_ahead
     sleep_delay_was = Delayed::Worker.sleep_delay
     min_reserve_interval_was = Delayed::Worker.min_reserve_interval
-
-    if Gem.loaded_specs['delayed'].version >= Gem::Version.new('1.0') && min_reserve_interval_was.zero?
-      raise "Min reserve interval should be nonzero in v1.0 release"
-    end
```
smudge, line 132: Deferring this new default indefinitely, as we have a "duty cycle" feature we're targeting to enable by default instead.

spec/helper.rb
```diff
@@ -151,6 +155,7 @@ class SingletonClass
     Delayed::Worker.read_ahead = read_ahead_was
     Delayed::Worker.sleep_delay = sleep_delay_was
     Delayed::Worker.min_reserve_interval = min_reserve_interval_was
+    Delayed.plugins = plugins_was
```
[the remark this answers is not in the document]
smudge: The line is already in an `ensure` block, closing this thread.

lib/delayed/lifecycle.rb
```diff
@@ -8,7 +8,7 @@ class Lifecycle
       perform: %i(worker job),
       error: %i(worker job),
       failure: %i(worker job),
```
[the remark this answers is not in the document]
smudge: We have started using GitHub releases as the primary changelog, and I intend to document it there.

lib/delayed/lifecycle.rb
```diff
@@ -8,7 +8,7 @@ class Lifecycle
...
-      thread: %i(worker job),
+      thread: [:worker],
```
smudge, line 11: This is the breaking change.

lib/delayed/worker.rb
```diff
@@ -142,17 +145,14 @@ def perform(job)
...
-
-        job.error = e
-        failed(job)
-        false # work failed
-      rescue Exception => e # rubocop:disable Lint/RescueException
-        self.class.lifecycle.run_callbacks(:error, self, job) { handle_failed_job(job, e) }
-        false # work failed
       end
+      true # did work
+    rescue DeserializationError => e
```
smudge, line 150: Most of these `rescue`s were pulled one level _out_ of the corresponding lifecycle's `do` + `end`. This means that if something inside a lifecycle hook fails, the "after" hook for that lifecycle will not trigger, which I believe is more in line with expectations. (e.g. if `perform` fails, there is no `after(:perform)` callback fired, but there _is_ an `after(:error)` callback.)

lib/delayed/worker.rb
```diff
@@ -101,12 +101,15 @@ def work_off(num = 100)
         pool = Concurrent::FixedThreadPool.new(jobs.length)
         jobs.each do |job|
           pool.post do
-            self.class.lifecycle.run_callbacks(:thread, self, job) do
+            self.class.lifecycle.run_callbacks(:thread, self) do
               success.increment if perform(job)
+            rescue DeserializationError => e
```
smudge, line 106: These two `rescue`s exist as a catch-all, just in case job cleanup is still possible at this point, but I was unable to actually exercise these codepaths because the ones in `perform` happen at the very end of that method now.

smudge: By clean up, I mean update run_at/attempts/failed_at/etc -- all the clean up work necessary to enable exponential backoff and prevent zombie jobs from building up and exhausting worker resources.

effron: For posterity, i think we can punt on this change for now to get this fix in, and then revisit!

PR #58, opened 2025-09-05
Drop support for Rails 5.2 & Ruby 2.7

Irving-Betterment:
**Summary of changes:**
This pull request drops support for the oldest Rails and Ruby versions, in order to streamline the supportability bracket. By doing so, It also cuts a minor version release.

### DIFF
--- a/lib/delayed/version.rb
+++ b/lib/delayed/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Delayed
-  VERSION = '2.0.0'
+  VERSION = '1.1.0'
 end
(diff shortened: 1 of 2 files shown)


lib/delayed/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Delayed
-  VERSION = '1.0.0'
```
argvniyx-enroute, line 4: I think we still stick to a minor bump when dropping a ruby version. See [this webvalve commit](https://github.com/Betterment/webvalve/commit/796dc0c6e4e2e318b5d6ea02c2895271de5f09b8), for example. Its parent is a commit dropping older ruby and rails versions.

smudge (approved):
domain LGTM && platform LGTM

Irving-Betterment:
@smudge can you please help me merging this approved PR?

PR #102, opened 2026-02-20
style: Add (passing) tests for regression coverage

smudge:
@medlefsen noticed that [ActiveJob states this in their docs](https://apidock.com/rails/ActiveJob/Enqueuing/ClassMethods/perform_later) w.r.t. `perform_later`:

> Returns an instance of the job class queued with arguments available in Job#arguments or false if the enqueue did not succeed.

Since silently returning `false` would appear to conflict with our preference for guaranteeing message delivery, I'm adding a couple regression test cases to make more explicit the current behavior, which presumably is specific to the AJ backend.

The `delayed` backend will not return `false` from `perform_later`, and will instead bubble out an exception.

/no-platform

### DIFF
--- a/spec/delayed/active_job_adapter_spec.rb
+++ b/spec/delayed/active_job_adapter_spec.rb
@@ -58,6 +58,22 @@ RSpec.describe Delayed::ActiveJobAdapter do
     end
   end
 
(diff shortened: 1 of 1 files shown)


samandmoore (approved):
domainlgtm platformlgtm

PR #103, opened 2026-04-21
feat: Implement enqueue_all

CelticMajora:
Implements `enqueue_all` in order to support Rails 7.1 perform_all_later

Key decisions:
- Maintain safety against `enqueue_after_transaction_commit`
- Adjusts shape of `enqueue` lifecycle hook
- Keep enqueueing logic isolated to `Delayed::Job` so that `Delayed::ActiveJobAdapter` can stay pure to job setup and post-enqueue object manipulation.

/domain @smudge @effron 
/platform @smudge @effron


lib/delayed/active_job_adapter.rb
```diff
@@ -14,8 +14,82 @@ def enqueue_at(job, timestamp)
...
+      # job N) rolls back jobs 0..N-1, matching the bulk path's all-or-nothing
+      # semantics.
+      Delayed::Job.transaction do
+        jobs.count do |job|
+          opts = job.scheduled_at ? { run_at: coerce_scheduled_at(job.scheduled_at) } : {}
+          _enqueue(job, opts)
+          job.successfully_enqueued = true
+          true
+        end
+      end
```
smudge: We may be better off `raise`ing in this case for now, as holding a long-running transaction around an N+1 insert will perform poorly under realistic conditions.

lib/delayed/active_job_adapter.rb
```diff
@@ -14,8 +14,82 @@ def enqueue_at(job, timestamp)
...
+        dj.hook(:enqueue)
+      end
+
+      # Replicate `before_save` hooks since insert_all bypasses callbacks.
+      dj.before_save_hooks
+      dj.attributes.compact
+    end
+
+    def coerce_scheduled_at(value)
+      value.is_a?(Numeric) ? Time.at(value) : value # rubocop:disable Rails/TimeZone
```
smudge, line 72: Does `job.scheduled_at` really need to be coerced? Seems reasonable to expect the implementation to construct a valid Time object.

CelticMajora: I wasn't sure if we needed to support old versions of ActiveJob where it was a numeric: https://apidock.com/rails/ActiveJob/Core/scheduled_at%3D

lib/delayed/active_job_adapter.rb
CelticMajora: Would it be preferred to have this delegate as if `enqueue_all` weren't implemented?

lib/delayed/active_job_adapter.rb
```diff
@@ -14,8 +14,82 @@ def enqueue_at(job, timestamp)
...
+        # Fall back to the per-job path when we can't take the bulk shortcut.
+        return enqueue_all_one_by_one(jobs) unless bulk_enqueue_supported?
+
+        rows = jobs.map { |job| build_insert_row(job) }
+        result = Delayed::Job.insert_all(rows, record_timestamps: true) # rubocop:disable Rails/SkipsModelValidations
+        ids = result.rows.map(&:first)
+
+        jobs.zip(ids) do |job, id|
+          job.provider_job_id = id
+          job.successfully_enqueued = true
```
smudge: Setting `successfully_enqueued` actually diverges from the singleton `_enqueue` method, and makes me wonder if we should be setting `successfully_enqueued` down there as well.

Additionally, the way that `_enqueue` handles the provider ID assignment has always felt a little vestigial, as the current implementation cannot _know_ the ID before it has persisted the job, at which point it would be pointlessly costly to issue an UPDATE to persist the ID into the ActiveJob attributes. As a result, the ActiveJob attributes always have a null `provider_job_id` upon read, so we make a point not to rely on it.

My thinking is that we could simplify this method significantly (_and_ support MySQL) if we skip handling `provider_job_id` after persistence.

(IMO, if we actually wanted to set it in the ActiveJob payload, we should generate a `uuidv7` up front and persist that both to the jobs table and the ActiveJob payload. But proper support starts to tug at other behaviors like `retry_on`, which can pass the same ActiveJob payload through multiple provider job rows, each with their own execution IDs.)

lib/delayed/active_job_adapter.rb
```diff
@@ -14,8 +14,82 @@ def enqueue_at(job, timestamp)
...
     end
 
+    # `perform_all_later` / `enqueue_all` is an ActiveJob 7.1+ feature.
+    # Earlier ActiveJob versions lack both the caller (`perform_all_later`)
+    # and the per-job `successfully_enqueued=` setter, so we gate only the
+    # public method — the private helpers are unconditional and just unused
+    # on <7.1.
+    if ActiveJob.gem_version >= Gem::Version.new('7.1')
+      def enqueue_all(jobs)
+        return 0 if jobs.empty?
```
smudge: It's not clear to me if `enqueue_all` is required to return anything: https://github.com/rails/rails/blob/eb126bb140127d3589ad9be093a845e24fc4f475/activejob/lib/active_job/enqueuing.rb#L19

app/models/delayed/job.rb
```diff
@@ -282,6 +282,11 @@ def attempts_alert?
       alert_attempts&.<= attempts
     end
 
+    def before_save_hooks
+      set_default_run_at
+      set_name
+    end
```
smudge: It would be nice to avoid expanding the public API of the Job model -- I understand wanting to avoid drift, but it also doesn't stop someone from adding a `before_save` and skipping callbacks again.

My thinking is that these should not be `before_save` calls at all and instead should be handled by the `enqueue` and `enqueue_all` methods. We could retain validations on this class to ensure that they are always set by the caller, but remove some of the magic so that the caller is always forced to call these two methods (or supply their own values) before saving.

smudge: Actually instead of validations, for our purposes, ensuring that we have appropriate DB constraints would be enough. (Otherwise we'd need to call `validate!` on each job while constructing the insert all payload, and that might not be worth the minor performance hit if we can get the guarantee from the DB itself.)

lib/delayed/active_job_adapter.rb
smudge: Not sure I follow. I think we should allow `enqueue_all` to proceed if it is supported, or raise an exception if the caller attempts to call it and it is not supported.

lib/delayed/active_job_adapter.rb
smudge: Ah, looks like that landed in 7.1, so we'd need to restrict our `activejob` dependency in turn. The activejob dependency is technically optional, but based on our [appraisals](https://github.com/Betterment/delayed/blob/main/Appraisals) we still support `>= 6.0`... 🤔 

I think we can consider changing support in a separate PR and keep the coercion in place here, if that makes sense to you.

smudge (changes requested):
TAFN -- Really excited for this! I left a few comments, but overall I'm looking to unify the implementation of `enqueue` and `enqueue_all` as much as possible. (Like, in theory, `enqueue(job)` could defined as `enqueue_all([job])`, and I'd like to see how close to that we can get!)

As it stands, my read is that the `enqueue_all` path makes an effort to replicate the lifecycle hooks of `Delayed::Backend::Base#enqueue` (e.g. using `JobPreparer`, calling the `before_save` methods, and running the `:enqueue` callback for each one), but the codepaths are not shared and would be subject to drift.

Additionally, pushing ourselves closer to a shared `enqueue_all(one_or_more_jobs)` implementation could also force some interesting questions about the necessity of some of these extra enqueue-time operations like the `:enqueue` lifecycle hook. I think there are some compelling arguments for dropping support for per-job operations/configurations that could instead be managed in bulk. (Like, if `:enqueue` as a lifecycle hook were to receive a list of jobs rather than a single job -- obv. a breaking change, but gets us the observability hook without the N+1 calls against each individual job object.)

I'm mostly spitballing in that last paragraph, but wanted to start a conversation, because we're in a territory where I think it would already be pretty reasonable to cut a new major version and break some older behaviors/assumptions.

lib/delayed/active_job_adapter.rb
CelticMajora: Other adapters I was looking at(ie: Solid Queue and Sidekiq) all returned the count, but it was unclear whether that was part of the contract.

CelticMajora:
Waiting on https://github.com/Betterment/delayed/pull/104 to move forward

lib/delayed/plugins/instrumentation.rb
```diff
@@ -2,9 +2,9 @@ module Delayed
   module Plugins
     class Instrumentation < Plugin
       callbacks do |lifecycle|
-        lifecycle.around(:enqueue) do |job, *args, &block|
```
CelticMajora, line 5: TODO: determine whether dropping `*args` here is needed

lib/delayed/plugins/instrumentation.rb
```diff
@@ -34,6 +34,16 @@ def self.active_support_notifications_tags(job)
           job: job,
         }
       end
+
+      def self.bulk_enqueue_tags(jobs)
```
CelticMajora, line 38: TODO, determine what is valuable here

lib/delayed/plugins/instrumentation.rb
CelticMajora: This is required, as the input otherwise tries to splat the jobs array

lib/delayed/active_job_adapter.rb
```diff
@@ -7,26 +7,63 @@ def enqueue_after_transaction_commit?
...
 
-    def _enqueue(job, opts = {})
-      if enqueue_after_transaction_commit_enabled?(job)
+    def assert_safe_to_enqueue!(jobs)
+      if jobs.any? { |job| enqueue_after_transaction_commit_enabled?(job) }
         raise UnsafeEnqueueError, "The ':delayed' ActiveJob adapter is not compatible with enqueue_after_transaction_commit"
       end
+      unless Delayed::Worker.delay_jobs == true
+        raise UnsafeEnqueueError, "The ':delayed' ActiveJob adapter is not compatible with delay_jobs false"
+      end
```
CelticMajora: Not confident on this restriction, but removing it makes the code quite a bit messier.

lib/delayed/plugins/instrumentation.rb
```diff
@@ -34,6 +34,16 @@ def self.active_support_notifications_tags(job)
...
+        {
+          count: jobs.size,
+          table: Delayed::Job.table_name,
```
smudge: The most accurate value here would be something like `jobs.map { |j| j.class.table_name }.uniq.sole`, but `jobs.first.class.table_name` might be a slightly cheaper proxy (assuming we enforce at the bulk enqueue API that all jobs go to the same table).

The reason I wouldn't assume it's always `Delayed::Job.table_name` is because it's possible (though not common) to create specialized job types for use with secondary DB connections or table names.

smudge: In fact, it might make sense to emit counts by class and let the caller decide what to do with those class names, e.g. `jobs.group_by { |j| j.class }.transform_values(&:count)`

smudge: @effron wdyt?

lib/delayed/plugins/instrumentation.rb
```diff
@@ -34,6 +34,16 @@ def self.active_support_notifications_tags(job)
...
+          database: Delayed::Job.database_name,
+          database_adapter: Delayed::Job.database_adapter_name,
```
smudge: Similar comment here w.r.t. relying on `jobs` to derive DB name and adatper.

lib/delayed/active_job_adapter.rb
```diff
@@ -7,26 +7,63 @@ def enqueue_after_transaction_commit?
...
+    def assign_provider_job_ids(jobs, result)
+      ids = result.rows.map(&:first)
+      jobs.zip(ids) { |job, id| job.provider_job_id = id }
+    end
 
-      Delayed::Job.enqueue(JobWrapper.new(job), opts).tap do |dj|
-        job.provider_job_id = dj.id
-      end
+    def mark_successfully_enqueued(jobs)
+      jobs.each { |job| job.successfully_enqueued = true if job.respond_to?(:successfully_enqueued=) }
```
smudge: Were you able to determine if this API is a necessary part of the contract?

lib/delayed/backend/base.rb
```diff
@@ -14,7 +14,7 @@ def enqueue(*args)
 
         def enqueue_job(options)
           new(options).tap do |job|
-            Delayed.lifecycle.run_callbacks(:enqueue, job) do
+            Delayed.lifecycle.run_callbacks(:enqueue, [job]) do
               job.hook(:enqueue)
```
smudge: This `job.hook(:enqueue)` is potentially going to call `enqueue` on every job in the list -- another thing to be mindful of if we're wanting to make enqueuing of many jobs fast by default.

smudge: Technically I think we also deprecated hook methods like this, so maybe it's time to finally remove support.

lib/delayed/active_job_adapter.rb
```diff
@@ -7,26 +7,63 @@ def enqueue_after_transaction_commit?
...
+
+      assert_safe_to_enqueue!(jobs)
+
+      Delayed.lifecycle.run_callbacks(:enqueue, jobs) do
+        now = Delayed::Job.db_time_now
+        rows = jobs.map { |job| build_insert_row(job, now) }
+        result = Delayed::Job.insert_all(rows) # rubocop:disable Rails/SkipsModelValidations
+        assign_provider_job_ids(jobs, result) if Delayed::Job.connection.supports_insert_returning?
+        mark_successfully_enqueued(jobs)
+      end
```
smudge: I'm wondering if we should move this entire body into `Delayed::Backend::Base` to live alongside `enqueue` and `enqueue_job`. (There are potentially further simplifications we could in turn make over there as well.)

lib/delayed/plugins/instrumentation.rb
smudge: It might also make sense to return IDs in this payload.

smudge (changes requested):
This is great! Some thoughts:

- Along with the change to the `:enqueue` lifecycle hook, I think it might make sense to finally remove the job-level `job.hook(<name>)` interface entirely. We had previously deprecated it, and `:enqueue` was already not supported for `ActiveJob` classes (due to a naming conflict). It's a breaking change (as are other changes in this PR, so we'll need a major version bump), but the question in my mind is mostly whether this will be a _hugely inconvenient_ change (in addition to being a breaking change).
- I'm not exactly sure what should go in the new `:enqueue` payload. We might want counts by job class (if we support polymorphic bulk enqueues), or at least a list of job IDs that were inserted into the DB, so that individual job logs can be traced back to when they were enqueued.
- Thinking that the bulk interface could be pushed all the way down to `Delayed::Job::Base` so that the ActiveJob adapter can really just be a thin wrapper around core enqueue functionality.

lib/delayed/plugins/instrumentation.rb
effron: i think having a metric per class is valuable, and we can use that as a proxy to know that all jobs with the same class have the same table etc?

effron: or, `job_name`, instead of class

lib/delayed/active_job_adapter.rb
CelticMajora: I think it is _sorta_ optional: https://github.com/rails/rails/pull/41191. `delayed` by definition makes `successfully_enqueued` sorta trivial due to our transactionality assurances. If the job wasn't successfully enqueued, there would be a `ActiveRecord::Rollback` to indicate

lib/delayed/plugins/instrumentation.rb
CelticMajora: Had a conversation with Harry and we landed on the tags just being:

```ruby
{
  jobs: jobs.map { |job| active_support_notifications_tags(job) },
}
```

This requires the enqueue_all to be shifted down to `Delayed::Backend::Base`(which is blocked by the two linked PRs below). Once those get in, I'll shift the enqueue_all and update the instrumentation

lib/delayed/active_job_adapter.rb
```diff
@@ -11,24 +11,67 @@ def enqueue(job)
...
 
     private
 
-    def _enqueue(job, opts = {})
+    def _enqueue(job)
+      assert_job_safe_to_enqueue!(job)
+      job_options = build_job_options(job)
+      delayed_job = Delayed::Job.enqueue_job(job_options)
+      perform_post_enqueue_assignments([job], [delayed_job])
+      job
```
smudge, line 38: Even if we want this to return the job that it just enqueued, I'm not quite following what else prevents us from folding the rest of the logic together:

```suggestion
      job.tap { |j| enqueue_all([job]) }
```

(LMK if I'm missing anything there!)

lib/delayed/lifecycle.rb
```diff
@@ -4,7 +4,7 @@ class InvalidCallback < RuntimeError; end
   class Lifecycle
     EVENTS = {
       execute: [nil],
-      enqueue: [:job],
+      enqueue: [:jobs],
```
smudge, line 7: This is a big enough change that we definitely need a major version bump. Might even be worth including it in this PR, so that if we ship anything else off of main we don't inadvertently include this in a minor/patch bump.

smudge (changes requested):
Thanks for this! All LGTM from an implementation standpoint -- had a question about some additional code cleanup above, but I don't think it's blocking. Also noted that we should bump the major version in this PR before merging it.

lib/delayed/active_job_adapter.rb
CelticMajora: I think the main thing is that `provider_job_id` will always be set when using Delayed::Job.enqueue_job directly, whereas using Delayed::Job.enqueue_all will assign `provider_job_id` only if the db supports insert all returning. What do you think?

smudge: I would rather apply a consistent behavior here, and `provider_job_id` being `nil` is (somewhat unfortunately) the best guarantee we can give.

This also applies to job round-trips. If I run `Delayed::Job.find(<id>)`, the `provider_job_id` will _always_ be `nil`, reflecting the catch-22 that we cannot know the PG-provided ID until after we persist the ActiveJob handler.

We should be sensitive to the possibility that this breaks something that had been relying on `provider_job_id` immediately after enqueue, but I'm comfortable with a breaking change here that removes the embedded `provider_job_id` behavior.

(I would prefer to extend the `ActiveJob::Core#provider_job_id` reader to behave differently on our job wrapper class, but that feels out of scope for this PR.)

smudge (approved):
LGTM if you're comfortable addressing [this](https://github.com/Betterment/delayed/pull/103#discussion_r3335987931) in a follow-up PR, or if you'd rather do it here I'm happy to circle back and re-review!

PR #104, opened 2026-04-22
refactor: Push run_at / name defaulting into JobPreparer

CelticMajora:
The before_save :set_default_run_at, :set_name callback is magical and doesn't actually prevent anyone from adding a new before_save that skips callbacks. Move the defaulting into Delayed::Backend::JobPreparer so both the single save path and any future bulk-insert path pick up defaults from the same place, and delete the before_save wiring along with the Job#set_name and Base#set_default_run_at helpers.

Add migrations to enforce the invariant at the DB level (with a COALESCE backfill for any legacy NULL rows). Direct Delayed::Job.create! / .save without supplying run_at and name is now a DB-level NOT NULL violation -- callers should route through Delayed::Job.enqueue (which runs JobPreparer).

Breaking change for anyone reaching past the documented enqueue API.


db/migrate/8_require_run_at_and_name.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class RequireRunAtAndName < ActiveRecord::Migration[6.0]
```
CelticMajora: main thing I'm iffy on is how to best build this "breaking" migration. In our traditional applications, we would handle the backfill, check constraint unvalidated, and check constraint validation in independent migrations but I'm not sure the best course of action for a public gem.

spec/delayed/__snapshots__/monitor_spec.rb.snap
```diff
@@ -51,9 +51,10 @@ GroupAggregate  (cost=...)
               ->  GroupAggregate  (cost=...)
                     Output: delayed_jobs.priority, delayed_jobs.queue, count(*), sum(CASE WHEN (delayed_jobs.run_at > '2025-11-10 17:20:13'::timestamp without time zone) THEN 1 ELSE 0 END), sum(CASE WHEN (delayed_jobs.attempts > 0) THEN 1 ELSE 0 END), min(CASE WHEN (delayed_jobs.run_at <= '2025-11-10 17:20:13'::timestamp without time zone) THEN delayed_jobs.run_at ELSE NULL::timestamp without time zone END)
                     Group Key: delayed_jobs.priority, delayed_jobs.queue
-                    ->  Sort  (cost=...)
+                    ->  Incremental Sort  (cost=...)
```
effron: do you know why there is a diff here?

db/migrate/8_require_run_at_and_name.rb
effron: is there a reason not to do it that way here?

spec/delayed/__snapshots__/monitor_spec.rb.snap
CelticMajora: According to claude it was due to the run_at column becoming non-nullable:

```
Both indexes involved have priority as their leading column:                                                                                                                                    
  - idx_delayed_jobs_live on (priority, run_at, queue, attempts) (live index, first failure)                                                                                                      
  - Legacy delayed_jobs_priority on (priority, run_at) (second failure, with :with_legacy_table_index)                                                                                            
                                                                                                                                                                                                
So Index (Only) Scan output is already sorted by priority, and the only ORDER BY remaining is (priority, queue). Incremental Sort exploits exactly this: priority is presorted, queue is sorted 
within each priority group. It is the more efficient plan.                                                                                                                                      
                                                                                                                                                                                                
What in your branch caused the flip                                                                                                                                                             
                                                                                                                                                                                              
The only schema-affecting change on this branch is db/migrate/8_require_run_at_and_name.rb, which runs change_column_null :delayed_jobs, :run_at, false. That is the load-bearing change here.  
                                                                                                                                                                                              
run_at is the second column in both indexes that appear in the failing plans. Once PG knows it is NOT NULL:                                                                                     
                                                                                                                                                                                              
  - Pathkey/ordering reasoning gets cleaner — no NULLs sentinel to worry about at the boundary of the leading-column "priority" groups.                                                           
  - The planner has slightly different selectivity / row-count estimates for the column.                                                                                                        
                                                                                                                                                                                                
On an essentially empty table (the SQL spec doesn't seed rows on Postgres — it just runs EXPLAIN), those small cost-estimate shifts are enough to flip the comparison so Incremental Sort beats 
Sort. 
```

If it really is based on an empty table though, I'm not sure how much weight to put into the change since that is an unrealistic state for the dj table.

db/migrate/8_require_run_at_and_name.rb
CelticMajora: Thought it was just a little weird to write it that way for rails 6 and non-postgres adapters, but it isn't too ugly. Switched over.

db/migrate/09_validate_run_at_and_name_not_null.rb
effron, line 1: do we care that this is no longer sorted correctly? we could zero pad the rest of the files so its sorted in the filesystem...

effron (dismissed):
one minor nit, otherwise this looks good! is the idea that this gets bundled into a major version bump?

db/migrate/09_validate_run_at_and_name_not_null.rb
CelticMajora: Good point, I can do a quick rename to make this sequence more clear.

CelticMajora:
> is the idea that this gets bundled into a major version bump?

Yeah, I'm thinking this will go into a major version bump that includes the `enqueue_all` support and any other associated breaking changes.

db/migrate/09_validate_run_at_and_name_not_null.rb
CelticMajora: VSCode automatically ordered things properly even without the change, so I hadn't even noticed that 😅

effron (dismissed):

db/migrate/08_backfill_null_run_at_and_name.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class BackfillNullRunAtAndName < ActiveRecord::Migration[6.0]
+  def up
+    Delayed::Job.where(run_at: nil).update_all('run_at = COALESCE(created_at, CURRENT_TIMESTAMP)')
```
smudge: This query may be more expensive than it is worth, as it would be very unlikely for a row to have a `NULL` `run_at` value, and I'd rather find out when the constraint fails to validate than paper over it.

db/migrate/08_backfill_null_run_at_and_name.rb
```diff
@@ -0,0 +1,12 @@
...
+    Delayed::Job.where(name: nil).update_all(name: 'Delayed::Job')
```
smudge: This would apply to any jobs that were created before `v1.2.0` and are still in the DB (e.g. they failed and haven't been cleaned up).

I'd be inclined to avoid prescribing this `update_all` statement. For one, it relies on an ActiveRecord being available in a migration, which isn't always the case. But even if I were confident it could be run in all downstream apps, I'd be inclined to put instructions in an `UPGRADING.md` file (which bundler can output if there's a reference in the gemspec) and treating this as a major version bump (e.g. 3.0) with breaking changes, and leaving the data backfill as an exercise for the app owner.

(Our recommendation can be to only ever upgrade 1 major version at a time, and do not expect jobs created 2 major versions ago to be guaranteed to work in the current major version.)

There's also the question of how to maintain a migration path from `delayed_job_active_record` (and its derivatives that use this schema). Until now, we've been basically 100% schema compatible.

lib/delayed/backend/job_preparer.rb
```diff
@@ -34,6 +36,24 @@ def set_priority
...
+        payload = options[:payload_object]
+        options[:name] =
+          if payload.respond_to?(:job_data)
+            payload.job_data['job_class']
+          elsif payload.respond_to?(:display_name)
+            payload.display_name
+          else
+            payload.class.name
+          end
+      end
```
smudge, line 55: We might be able to simplify this by relying on the fact that `JobWrapper` defines a `display_name`:

https://github.com/Betterment/delayed/blob/00d575c23b918cd79a174be7617c0847a40ab9cb/lib/delayed/job_wrapper.rb#L19

And we could make `Delayed::JobBase` also define its own `display_name` as `self.class.name`.

Which then gives us:

```ruby
options[:name] = options[:payload_object].display_name
```

WDYT?

lib/delayed/backend/base.rb
```diff
@@ -57,14 +57,6 @@ def failed?
 
       ParseObjectFromYaml = %r{!ruby/\w+:([^\s]+)} # rubocop:disable Naming/ConstantName
 
-      def name
-        if self.class.column_names.include?('name')
-          super || display_name
-        else
-          display_name # [feat:NameColumn] remove fallback once the "name" column is required.
-        end
-      end
```
smudge, line 66: Re: the backwards compatibility question I asked [here](https://github.com/Betterment/delayed/pull/104/changes#r3170573389), one transitional approach might be to go from three options:

- name is present
- name is NULL
- name column does not exist

to two options:

- name is present (non-NULL constraint on column)
- name column does not exist

Do you think it would be significantly more complex to support the latter combination? (It would allow us to further kick the can down the road on needing a more complicated migration path for schemas that started on `delayed_job_active_record`.)

lib/delayed/backend/job_preparer.rb
CelticMajora: Tried putting a top level display_name on `Delayed::Backend::Base`, but when the name is being set in the preparer, the payload_object is still the raw job object rather than one that has been wrapped in a delayed job. This means we have no assurances that it will implement a display_name.

CelticMajora:
Adjusted such that the code changes here won't break if the consumer still lacks the name column on their dj table.

lib/delayed/backend/job_preparer.rb
```diff
@@ -34,6 +36,24 @@ def set_priority
...
+
+        payload = options[:payload_object]
+        options[:name] =
+          if payload.respond_to?(:job_data)
+            payload.job_data['job_class']
+          elsif payload.respond_to?(:display_name)
+            payload.display_name
+          else
+            payload.class.name
+          end
```
CelticMajora, line 54: Currently not DRY with `display_name` backup on `Delayed::Backend::Base`, but not sure if there is a way around that while we still allow the name column to not exist.

lib/delayed/backend/base.rb
```diff
@@ -58,8 +62,8 @@ def failed?
       ParseObjectFromYaml = %r{!ruby/\w+:([^\s]+)} # rubocop:disable Naming/ConstantName
 
       def name
-        if self.class.column_names.include?('name')
-          super || display_name
+        if self.class.name_assignable?
+          super
```
smudge: It may make sense to keep the fallback here so that the migration path of `no column => nullable column => non-nullable column` still works with the latest version of the gem.

```suggestion
          super || display_name
```

spec/delayed/job_spec.rb
```diff
@@ -493,37 +490,12 @@ def create_job(opts = {})
...
-      end
-
-      it 'parses from handler on deserialization error' do
-        job = Story.create(text: '...').delay.text
-        job.payload_object.object.destroy
-        job.save!
-        job.update_column(:name, nil) # rubocop:disable Rails/SkipsModelValidations
-        expect(job.reload.name).to eq('Delayed::PerformableMethod')
-      end
-    end
```
smudge, line 525: To prove that [this](https://github.com/Betterment/delayed/pull/104/changes#r3183748431) intermediate falllback still works, you could keep this context and add an `around` that rolls back and then (in an `ensure`) rolls forward the migrations that set the column to non-null.

lib/delayed/backend/job_preparer.rb
smudge: In the active job case, is `payload` a `Delayed::JobWrapper` already by this point, or does that get wrapped later?

I'm thinking that you might at least be able to remove the branch that checks for `job_data`, because we can guarantee that `Delayed::JobWrapper` defines a `display_name`.

smudge (changes requested):
TAFN -- a couple more questions, but this feels very close!

lib/delayed/backend/job_preparer.rb
CelticMajora: That kinda breaks down with `ActiveJob::QueueAdapters::DelayedJobAdapter::JobWrapper` usages. Not sure if that is a realistic production scenario, but we have specs testing against that. Since that job wrapper defines `display_name` [differently](https://github.com/collectiveidea/delayed_job/blob/eaef62d844ea980433b807a496136bbd0a6ea625/lib/active_job/queue_adapters/delayed_job_adapter.rb#L30-L36), if we want consistent behavior I think we need to rely on `job_data`.

smudge: Oh, good point. I guess we do technically support that queue adapter (though we warn against it).

Happy to leave-as is, but this is definitely worth revisiting if we ever fully block that usage.

smudge (approved):

PR #106, opened 2026-05-18
refactor: deprecate :enqueue job hook

CelticMajora:
Remove the job.hook(:enqueue) invocation from enqueue_job so payloads can no longer mutate or observe the job at enqueue time. The hook method now short-circuits on :enqueue with a deprecation warning for any direct callers, and enqueue_job emits its own deprecation warning (naming the payload class) when a non-ActiveJob payload still defines :enqueue, so users discover the broken hook at enqueue time rather than silently.

Drop tests and sample jobs that existed only to exercise the hook (EnqueueJobMod, CallbackJob#enqueue, PerformableMethod's enqueue delegation spec) and remove the 'enqueue' message expectations from the remaining callback tests. Add specs covering the new warn paths, including the regression guard that ActiveJob JobWrapper payloads do not trigger the deprecation warning.

cc @effron @smudge

### DIFF
--- a/lib/delayed/backend/base.rb
+++ b/lib/delayed/backend/base.rb
@@ -14,8 +14,8 @@ module Delayed
 
         def enqueue_job(options)
           new(options).tap do |job|
+            warn_deprecated_enqueue_hook(job.payload_object)
             Delayed.lifecycle.run_callbacks(:enqueue, job) do
-              job.hook(:enqueue)
               Delayed::Worker.delay_job?(job) ? job.save : job.invoke_job
             end
           end
@@ -44,6 +44,15 @@ module Delayed
 
         def name_assignable?
           column_names.include?('name')
+        end
+
+        private
+
+        def warn_deprecated_enqueue_hook(payload)
+          return if payload.is_a?(Delayed::JobWrapper)
+          return unless payload.respond_to?(:enqueue)
+
+          warn "[DEPRECATION] :enqueue hook on #{payload.class} is deprecated and is no longer invoked"
         end
       end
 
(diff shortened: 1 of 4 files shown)


effron (dismissed):

lib/delayed/backend/base.rb
```diff
@@ -109,9 +118,12 @@ def unlock
 
       def hook(name, *args)
         if payload_object.respond_to?(name)
-          if payload_object.is_a?(Delayed::JobWrapper)
-            return if name == :enqueue # this callback is not supported due to method naming conflicts.
+          if name == :enqueue
+            warn '[DEPRECATION] :enqueue hook is deprecated'
+            return
+          end
```
CelticMajora, line 123: In theory this shouldn't come up, but `hook` is part of the public interface so in theory consumers could be calling it directly with `:enqueue`

lib/delayed/backend/base.rb
```diff
@@ -109,9 +118,12 @@ def unlock
 
       def hook(name, *args)
         if payload_object.respond_to?(name)
-          if payload_object.is_a?(Delayed::JobWrapper)
-            return if name == :enqueue # this callback is not supported due to method naming conflicts.
+          if name == :enqueue
+            warn '[DEPRECATION] :enqueue hook is deprecated'
+            return
```
smudge: Deprecation implies that it's still functioning but will soon be removed -- thoughts on a `raise` instead of a `warn`?

I'd like to avoid a situation where callbacks silently stop triggering (even if "silent" comes with a log line - which we might catch with `uncruft`, but that's unique to our app config).

lib/delayed/backend/base.rb
```diff
@@ -45,6 +45,15 @@ def work_off(num = 100)
...
           column_names.include?('name')
         end
+
+        private
+
+        def warn_deprecated_enqueue_hook(payload)
+          return if payload.is_a?(Delayed::JobWrapper)
+          return unless payload.respond_to?(:enqueue)
+
+          warn "[DEPRECATION] :enqueue hook on #{payload.class} is deprecated and is no longer invoked"
```
smudge: Similar note here - a `raise` would be preferable to a log line.

smudge: (and I wouldn't call it deprecated if it is already removed/disabled)

lib/delayed/backend/base.rb
CelticMajora: Addressed 👍

lib/delayed/backend/base.rb
CelticMajora: Addressed 👍

smudge (approved):

PR #107, opened 2026-05-19
refactor: Drop support for Delayed::Worker.delay_jobs being proc

CelticMajora:
After talking more with @effron and @smudge, this plan will replace https://github.com/Betterment/delayed/pull/105. In the short term, this makes the breaking change less breaking so that we can prepare for a full removal of `Delayed::Worker.delay_jobs` in the future.

### DIFF
--- a/lib/delayed/backend/base.rb
+++ b/lib/delayed/backend/base.rb
@@ -16,7 +16,7 @@ module Delayed
           new(options).tap do |job|
             Delayed.lifecycle.run_callbacks(:enqueue, job) do
               job.hook(:enqueue)
-              Delayed::Worker.delay_job? ? job.save : job.invoke_job
+              Delayed::Worker.delay_jobs ? job.save : job.invoke_job
             end
           end
         end
--- a/lib/delayed/backend/base.rb
+++ b/lib/delayed/backend/base.rb
@@ -16,7 +16,7 @@ module Delayed
           new(options).tap do |job|
             Delayed.lifecycle.run_callbacks(:enqueue, job) do
               job.hook(:enqueue)
-              Delayed::Worker.delay_job?(job) ? job.save : job.invoke_job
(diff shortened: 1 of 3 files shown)


lib/delayed/worker.rb
```diff
@@ -37,9 +37,14 @@ class << self
...
-    def self.delay_job?(job)
+    def self.delay_job?
       if delay_jobs.is_a?(Proc)
-        delay_jobs.arity == 1 ? delay_jobs.call(job) : delay_jobs.call
+        if delay_jobs.arity.positive?
+          warn '[DEPRECATION] delay_jobs proc arity 1 support is deprecated and behaves as delay_jobs=true'
+          true
+        else
+          delay_jobs.call
+        end
```
smudge: (from offline convo) I think we can drop support for procs entirely.

CelticMajora: Updated!

effron (dismissed):

lib/delayed/backend/base.rb
```diff
@@ -16,7 +16,7 @@ def enqueue_job(options)
           new(options).tap do |job|
             Delayed.lifecycle.run_callbacks(:enqueue, job) do
               job.hook(:enqueue)
-              Delayed::Worker.delay_job?(job) ? job.save : job.invoke_job
+              Delayed::Worker.delay_jobs ? job.save : job.invoke_job
```
smudge, line 21: I think it would make sense to raise if  `delay_jobs.is_a?(Proc)` is true, so that anyone who was relying on a proc will not suddenly have all cases become "truthy" and call `save` instead of `invoke_job`.

CelticMajora: Addressed 👍

smudge (approved):

PR #109, opened 2026-05-28
warn if max_claims is greater than or equal to connection pool size

bKP451:
## The Problem

Every job thread needs a database connection to do its work. But the worker process itself also needs one connection just to keep the lights on — polling for new jobs, locking them, and cleaning up when done.

If `max_claims` is set as high as (or higher than) the connection pool size, there aren't enough connections to go around:

```
MAX_CLAIMS = 5, POOL_SIZE = 5

main worker housekeeping ─🔑          ← needs 1
Job 1 ─🔑                        \
Job 2 ─🔑                         \
Job 3 ─🔑                          ├─ needs 5
Job 4 ─🔑                         /
Job 5 ─❌ (no connection left!)  /

6 requests  ▶  5 connections  ▶  somebody loses
```

Job 5 waits for a free connection, times out, and raises `ActiveRecord::ConnectionTimeoutError`. This is silent until it happens under load in production — with no obvious hint that the root cause is a misconfigured pool size.

## What This PR Does

Adds a one-time check at worker startup. If `max_claims >= connection_pool.size`, the worker logs a clear `warn`-level message telling you exactly what to fix before any jobs run.

The fix is simple — reserve one connection for the worker itself:

```ruby
# config/initializers/delayed_job.rb
Delayed::Worker.max_claims = ActiveRecord::Base.connection_pool.size - 1
```

More generally, if your jobs each need `N` connections and you want `M` concurrent threads, size your pool to at least `N * (M + 1)`. For most apps `N=1`, so `pool_size >= max_claims + 1` is the rule of thumb.

## What Changed

- `Worker#start` now calls `check_connection_pool_config!` before the run loop
- `check_connection_pool_config!` warns at startup when `max_claims >= connection_pool.size`, with the suggested value to set
- README documents the constraint with the formula and an example initializer
- If connection pool introspection raises for any reason, the check is silently skipped — no impact on correctly configured workers

### DIFF
--- a/lib/delayed/worker.rb
+++ b/lib/delayed/worker.rb
@@ -88,20 +88,14 @@ module Delayed
 
       while total < num
         start = clock_time
-        say "Attempting to reserve up to #{num - total} job(s)", 'debug'
         jobs = reserve_jobs
-        say 'No jobs reserved; exiting work_off loop', 'debug' if jobs.empty?
         break if jobs.empty?
 
         total += jobs.length
-        say "Reserved #{jobs.length} job(s); dispatching batch", 'debug'
-        pool = Concurrent::FixedThreadPool.new(thread_pool_size(jobs.length))
+        pool = Concurrent::FixedThreadPool.new(jobs.length)
         jobs.each do |job|
-          job_say job, 'Queued for thread execution', 'debug'
           pool.post do
-            thread_started = false
             self.class.lifecycle.run_callbacks(:thread, self) do
-              thread_started = true
               success.increment if perform(job)
             rescue DeserializationError => e
               handle_unrecoverable_error(job, e)
@@ -109,19 +103,16 @@ module Delayed
               handle_erroring_job(job, e)
             end
           rescue Exception => e # rubocop:disable Lint/RescueException
-            handle_thread_error(job, e, thread_started)
+            say "Job thread crashed with #{e.class.name}: #{e.message}", 'error'
           end
         end
 
-        say 'Waiting for worker threads to finish', 'debug'
         pool.shutdown
         pool.wait_for_termination
-        say "Batch finished with #{success.value} successful job(s) out of #{total} attempted so far", 'debug'
 
         break if stop? # leave if we're exiting
 
         elapsed = clock_time - start
-        say format('Batch elapsed %.4f seconds', elapsed), 'debug'
         interruptable_sleep(self.class.min_reserve_interval - elapsed)
       end
 
@@ -242,17 +233,6 @@ module Delayed
 
     def reload!
       Rails.application.reloader.reload! if defined?(Rails.application.reloader) && Rails.application.reloader.check!
-    end
-
-    def thread_pool_size(job_count)
-      return job_count unless Delayed::Job.respond_to?(:connection_pool)
-
-      pool_size = Delayed::Job.connection_pool.size
-      return job_count unless pool_size
-
-      [job_count, [pool_size - 1, 1].max].min
-    rescue StandardError
-      job_count
     end
 
     def clock_time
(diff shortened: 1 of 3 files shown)


lib/delayed/worker.rb
```diff
@@ -235,6 +254,17 @@ def reload!
...
+    def thread_pool_size(job_count)
+      return job_count unless Delayed::Job.respond_to?(:connection_pool)
+
+      pool_size = Delayed::Job.connection_pool.size
+      return job_count unless pool_size
+
+      [job_count, [pool_size - 1, 1].max].min
+    rescue StandardError
+      job_count
+    end
```
smudge: There may be a way to more proactively (e.g. during worker boot/initialization) establish if `Delayed::Worker.max_claims > Delayed::Job.connection_pool.size` rather than performing this `thread_pool_size` logic on every pickup loop. (My understanding is that `Delayed::Job.connection_pool.size` is informed by the pool size config in `database.yml`, and should not change once the app has loaded.)

The current pickup strategy is also intended to avoid picking up more work than the worker can immediately begin working off (to avoid holding unworked jobs in memory), so it may make sense to `raise` or `warn` up front (again, during boot / worker initialization) if a misconfiguration is detected.

smudge (changes requested):
Thanks for this PR!

I know you've marked it WIP/Temp, but I have also noticed that it's possible to trigger `ActiveRecord::ConnectionTimeoutError` if the pool size is lower than `max_claims`, so this fix definitely makes sense to me and I would be very open to addressing it. 👍 

For the sake of reviewing changes independently, I think it would make sense to split the observability/logging changes out into a separate PR, and keep this PR focused on the thread pool / connection pool / max claims reconciliation. (LMK if that doesn't make sense!)

lib/delayed/worker.rb
bKP451: @smudge,

Yeah it is a great way to `raise` on `warn` upfront during app initializer ( _config/initializers/delayed.rb_ ) for 
`Delayed::Worker.max_claims > Delayed::Job.connection_pool.size`


On our project, our actual problem was  max_claims being equal to pool_size. On investigation we found that main worker ( i.e server ) also needs to connect to DB to keep track of threads, locking, unlocking, polling etc. Therefor a worker thread will throw `ActiveRecord::ConnectionTimeoutError` if it cannot get DB connection after awaiting for a checkout_time ( i.e `ActiveRecord::Base.connection_pool.checkout_timeout` ). For long running jobs,  DB connection won't get free for that worker thread and exception is raised


`Job thread crashed with ActiveRecord::ConnectionTimeoutError: could not obtain a connection from the pool within 5.000 seconds (waited 5.001 seconds); all pooled connections were in use`
Here 5.000 seconds is checkout_time



**Problem illustration**

```
   MAX_CLAIMS = 5, POOL_SIZE = 5


   main worker(server) housekeeping ─🔑          ← needs 1
   Job 1 ─🔑                        \
   Job 2 ─🔑                         \
   Job 3 ─🔑                          ├─ needs 5
   Job 4 ─🔑                         /
   Job 5 ─❌ (no key left!)         /

   6 requests  ▶  5 keys  ▶  somebody loses
```

We solved it by `reducing concurrent worker threads to pool minus 1`

Override MAX_CLAIMS at config/initializers/delayed.rb


```ruby
# Reserve 1 DB connection for the worker's own housekeeping (polling + locking jobs).
db_pool_size = ActiveRecord::Base.connection_pool.size
Delayed::Worker.max_claims = [db_pool_size - 1, 1].max
```


WDYT ?

smudge: Yes, that's a good callout -- we always include a buffer in our connection pool size (both for web and worker counts). We also sometimes need more than 1 connection per thread, but the same general rule applies there too -- basically, if your code generally needs N connections, and your max claims is M, you want N * (M + 1) connections. The most common case is `N=1` though, and I think that would be easy enough to detect with `>=` (rather than the `>` I had originally proposed):

```
Delayed::Worker.max_claims >= Delayed::Job.connection_pool.size
```

PR #113, opened 2026-06-02
fix: Assert ActiveJobs can not be passed directly to DJ enqueue methods without going through the adapter

CelticMajora:
Introduces new assertion that enforces that an ActiveJob cannot be passed into DJ enqueue methods, indicating to consumers that `perform_later` should be used instead in order to pipe through the adapter.

### DIFF
--- a/lib/delayed/backend/base.rb
+++ b/lib/delayed/backend/base.rb
@@ -82,7 +82,7 @@ module Delayed
         end
 
         def assert_no_enqueue_hook!(payload)
-          return if payload.is_a?(Delayed::JobWrapper)
+          return if payload.is_a?(Delayed::JobWrapper) || payload.is_a?(ActiveJob::Base)
           return unless payload.respond_to?(:enqueue)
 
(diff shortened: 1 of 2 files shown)


lib/delayed/backend/base.rb
```diff
@@ -82,7 +82,7 @@ def bulk_insert_all(jobs)
         end
 
         def assert_no_enqueue_hook!(payload)
-          return if payload.is_a?(Delayed::JobWrapper)
+          return if payload.is_a?(Delayed::JobWrapper) || payload.is_a?(ActiveJob::Base)
```
smudge: I've personally experienced some odd behaviors when `ActiveJob::Base` classes are passed directly into `Delayed::Job.enqueue`. I suspect it may actually be a kind of undefined behavior, since it's not the intended execution path.

Would it make sense to `raise` proactively instead with a message that indicates `perform_later` must be called instead?

smudge (approved):

PR #114, opened 2026-06-15
fix: max_age metric excludes locked jobs

effron:
This pr https://github.com/Betterment/delayed/pull/96 made a change to how max_age is calculated, where it started including the run_at's of jobs that were currently locked. This made our max_age monitor go off even when we had a single long-running job in the queue, instead of showing us when the queue was backed up.

This restores the previous behavior where max_age only includes jobs that are pending, and not actively being worked off.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    delayed (3.0.0)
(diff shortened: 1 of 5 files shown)


CelticMajora (approved):
lgtm 🎉

PR #115, opened 2026-07-08
Define named priority methods as real methods

frewsxcv:
## What

Replaces `method_missing` dispatch for named priority lookups with real method definitions:

- **Class methods** — `Delayed::Priority.eventual`, `.interactive`, etc.
- **Instance predicates** — `priority.eventual?`, `priority.interactive?`, etc.

Methods are defined when the class loads (for the default names) and are removed/redefined whenever `Delayed::Priority.names=` is assigned. Bodies still resolve through `names_to_priority` at call time, so `assign_at_midpoint` behavior is unchanged.

## Why

`method_missing` methods are invisible to runtime reflection, so RDoc and Sorbet's tapioca can't see them. In practice this means every Sorbet-adopting app hits errors like:

```
Method eventual does not exist on T.class_of(Delayed::Priority) https://srb.help/7003
```

on every `queue_with_priority Delayed::Priority.eventual` call site, and has to hand-maintain a shim RBI. (In Betterment's retail monorepo this single pattern is currently the only thing blocking ~95 files from `typed: true`.) With real definitions, `tapioca gems` picks the methods up automatically:

```ruby
Delayed::Priority.methods(false)
# => [..., :eventual, :interactive, :reporting, :user_visible, ...]
```

## Behavior notes

- **Pre-existing methods are never overridden.** Under `method_missing`, a real method always won (e.g. a hypothetical name of `superclass` would never shadow `Class#superclass`); the new definitions skip any name/predicate that already resolves, preserving that exactly. Covered by a new spec.
- **`respond_to?` semantics are unchanged**, including `respond_to?(:interactive) == false` after names are customized (existing specs cover this).
- Apps that customize `names=` in an initializer still won't see their custom names in a tapioca-generated gem RBI (tapioca doesn't boot the app) — same as today; they keep their shims.

## Also

Fixes a pre-existing test-order leak in `priority_spec.rb`: the `around` hook reset `names` and `alerts` but not `assign_at_midpoint`, so any spec running after a midpoint example observed midpoint-shifted values. Reproducible on `main` with:

```
bundle exec rspec --order defined spec/delayed/priority_spec.rb:158 spec/delayed/active_job_adapter_spec.rb:149
```

## Testing

- `spec/delayed/priority_spec.rb` extended: real-method introspection, method cleanup on `names=` reassignment, and collision safety.
- Full suite run locally on sqlite across multiple seeds; only failures are pre-existing on `main` in my local environment (Rails 8.0.2 snapshot/error-class drift, unrelated to this change).

### DIFF
--- a/lib/delayed/priority.rb
+++ b/lib/delayed/priority.rb
@@ -70,11 +70,12 @@ module Delayed
       def names=(names)
         raise "must include a name for priority >= 0" if names && !names.value?(0)
 
+        remove_named_priority_methods
         @ranges = nil
         @alerts = nil
         @names_to_priority = nil
         @names = names&.sort_by(&:last)&.to_h&.transform_values { |v| new(v) }
-        redefine_named_priority_methods
+        define_named_priority_methods
       end
 
       def alerts=(alerts)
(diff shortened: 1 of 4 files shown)


lib/delayed/priority.rb
```diff
@@ -124,16 +125,30 @@ def midpoint(low, high)
...
-        names_to_priority.key?(method_name) || super
+      # Defines a class method (e.g. `Priority.eventual`) and an instance predicate (e.g.
+      # `priority.eventual?`) per name, as real, introspectable methods (visible to `methods`,
+      # RDoc, and Sorbet's tapioca, unlike `method_missing` dispatch). Pre-existing methods
+      # (e.g. a name of `superclass`) are never overridden.
+      def redefine_named_priority_methods
+        @named_priority_class_methods&.each { |m| singleton_class.send(:remove_method, m) }
+        @named_priority_predicates&.each { |m| remove_method(m) }
+        @named_priority_class_methods = names.keys.filter_map { |name| define_priority_class_method(name) }
+        @named_priority_predicates = names.keys.filter_map { |name| define_priority_predicate(name) }
```
rzane: Can this be done without introducing additional pieces of state?

smudge: We do still want to call `remove_method` for any previously-specified names, but I agree I think this logic could be implemented with `@names` as the only retained state.

frewsxcv: @rzane @smudge how does this look? https://github.com/Betterment/delayed/pull/115/changes/b6e303e4cff50cf22a21af640974d1e9f414eef4

rzane (dismissed):

smudge (approved):

PR #117, opened 2026-07-31
Monitor by tag_columns (defaults with no tag_columns)

BetterBoyer:
**What**

All `delayed.job.*` monitor metrics can now be grouped by — and tagged with — a configurable set of job columns, in addition to `priority` and `queue`. This is opt-in: `Delayed::Monitor.tag_columns` defaults to `[]`, and with the default the emitted payloads are identical to before.

Any column on the jobs table can be tagged, including custom columns an application adds (`table`/`database`/`database_adapter` tags elided below):

```ruby
# example with jobs table having the `name` column
Delayed::Monitor.tag_columns = %i(name)

# delayed.job.count emits one series per (priority, queue, name), plus an
# untagged zero baseline per (priority, queue) — here, one enqueued job:
{ priority: 'interactive',  queue: 'default', name: 'SimpleJob', value: 1 }
{ priority: 'interactive',  queue: 'default', value: 0 }
{ priority: 'user_visible', queue: 'default', value: 0 }
{ priority: 'eventual',     queue: 'default', value: 0 }
{ priority: 'reporting',    queue: 'default', value: 0 }
```

```ruby
# example with jobs table having both `name` and `owner` columns
Delayed::Monitor.tag_columns = %i(name owner)

# both columns land in the same GROUP BY, so each series carries both tags —
# one series per (priority, queue, name, owner). Here: 4 interactive SimpleJob
# jobs with no owner, plus one owned by team_a and one by team_b.
{ priority: 'interactive',  queue: 'default', name: 'SimpleJob', owner: 'unset',  value: 4 }
{ priority: 'interactive',  queue: 'default', name: 'SimpleJob', owner: 'team_a', value: 1 }
{ priority: 'interactive',  queue: 'default', name: 'SimpleJob', owner: 'team_b', value: 1 }
{ priority: 'interactive',  queue: 'default', value: 0 }
{ priority: 'user_visible', queue: 'default', value: 0 }
{ priority: 'eventual',     queue: 'default', value: 0 }
{ priority: 'reporting',    queue: 'default', value: 0 }
```

**Why**

When `delayed.job.max_age` alerts, a common first question is "which job is stuck?" (and often "whose is it?"). Previously that required querying the database by hand. Now it's a `max by {queue, priority, name}` away in the metrics provider when `tag_columns` is configured with `:name`. Since similar questions can arise from other custom columns, `tag_columns` can be configured for multiple columns.

**How**

- `grouped_query` accepts `extra_group_columns`, threaded into both the inner and outer `GROUP BY`s, so the query *count* is unchanged — still one query each for `live`/`pending`/`failed` per run. The query *cost* is not. No existing index contains a tag column (`idx_delayed_jobs_live` is `(priority, run_at, locked_at, queue, attempts)` partial on `failed_at IS NULL`; `idx_delayed_jobs_failed` is `(priority, queue)`), so tagging costs the monitor its covering-index reads and adds a sort. The plan snapshots taken while `%i(name)` was still the default (34c6105) measure this against `main`:
  - **postgres:** the index-only scans behind `count` (both the live and failed halves) become plain index scans — a heap fetch per matching row — each with an added sort
  - **mysql:** the failed-jobs query drops from a covering index range scan to a full table scan; the live and pending queries lose covering-index lookups
  - **sqlite:** same index scans, plus a temp B-tree for the failed query's inner `GROUP BY`
  These run every `sleep_delay` (60s by default), and plan choices shift with table size, so check the plans against a production-sized jobs table before enabling a tag column. Documented in the README.
- `Delayed::Monitor.tag_columns=` validates columns against the schema and raises `ArgumentError` at assignment (i.e. at boot when set in an initializer) rather than failing later in the monitor run loop
  **->-> Note this means setting it in an initializer requires a DB connection at boot <-<-**
- `NULL` tag values are reported as `'unset'` (e.g. jobs enqueued before the column existed)

**Behavioral notes / compatibility**

- No new event names, and no removed ones
  - Existing subscribers keep working, and with the default `[]` the payloads and query shapes are unchanged
  - Once configured, a metric's cardinality is multiplicative across tag columns (priority × queue × name × owner) and scales with the number of distinct values; untagged zero baselines are still emitted for every (`priority`, configured `queue`) regardless of whether tagged series exist
  - Dashboards aggregating these gauges should `sum by` / `max by` accordingly
- The `README.md` documents the behavior, the cardinality and index tradeoffs, and a three-deploy rollout guide for adding a custom tag column. It also now documents `delayed.job.locked_count`, which was previously undocumented.

**Testing**

- Full suite passes, with new coverage for: multiple names per (`priority`, `queue`), `'unset'` reporting, custom columns, and assignment-time validation. The pre-existing examples continue to cover the empty-`tag_columns` default.
- SQL/query-plan snapshots are **unchanged** across sqlite/mysql/postgres, confirming the default configuration emits the same queries as before. Note that this leaves no pinned plan for the tagged `GROUP BY` — the tagged plans quoted above live only in this PR's history (34c6105), since the `describe 'SQL'` block reverted to the untagged default in a0dba18.

**Other notes**

Initially, the idea was to just add the ability to tag _ONLY_ by name. This is captured in the first commit. Then the thinking evolved to enable fully customizable (metric, tag column) configurations. This was rejected as it makes more sense to have each tag column emitted across all metrics which exist instead as it seemed unlikely we would want to configure `max_age_by_name` and not, for example, `failed_count_by_name`. The final commit is where this effort landed so that each tagged column is emitted against each metric instead. Noting this evolution in case it (notably the second commit) helps highlight future capabilities that might be desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,10 +450,15 @@ Delayed::Monitor.tag_columns = %i(name owner)
 
 A few behavioral notes:
 
-- Each tag column is added to the monitor's `GROUP BY`, and the generated indexes (e.g.
-  `idx_delayed_jobs_live`) cover only `priority` and `queue`. Grouping by anything else can push
-  these queries off their index and into significantly more expensive scans, so evaluate any tag
-  column by checking the query plans against a production-sized jobs table.
+- Each tag column is added to the monitor's `GROUP BY`, and no generated index contains a tag
+  column: `idx_delayed_jobs_live` covers `(priority, run_at, locked_at, queue, attempts)` and
+  `idx_delayed_jobs_failed` covers `(priority, queue)`. Grouping by a column outside these indexes
+  costs the monitor its covering-index reads and adds a sort. On PostgreSQL, the index-only scans
+  behind `delayed.job.count` become plain index scans — a heap fetch per matching row. On MySQL,
+  the failed-jobs query drops from a covering index range scan to a full table scan. These are
+  queries the monitor re-runs every `sleep_delay` (60 seconds by default), and plan choices shift
+  with table size, so check the plans against a production-sized jobs table before enabling a tag
+  column.
 - Rows whose value was never populated for a tagged column are reported under the value `'unset'`
   (e.g. jobs enqueued before the `name` column existed, mid-upgrade).
 - Configured columns must exist on the jobs table: assigning a missing column to `tag_columns`
--- a/README.md
+++ b/README.md
@@ -434,8 +434,26 @@ An additional _experimental_ metric is available, intended for use with applicat
 
 - **delayed.job.alert_age_percent** - the _percent_ to which the oldest job has reached the "age alert" threshold. (See the [Alerting Threshholds](#priority-based-alerting-threshholds) section above.)
 
-All of these events may be subscribed to via a single regular expression (again, in your application
-config or in an initializer):
+If your jobs table has a `name` column (see [Database Setup](#database-setup)), one additional
+event will be emitted, grouped by priority name, queue name, **and job name**:
+
+- **delayed.job.max_age_by_name** - the age of the oldest run_at value, per job name (excludes failed jobs)
+
+This answers "_which_ job is stuck?" when `delayed.job.max_age` alerts. A few behavioral notes:
+
+- Unlike the other metrics, it is not zero-filled: job names cannot be enumerated in advance, so a
+  (priority, queue, name) series is only emitted while matching jobs are present in the queue.
+- Jobs enqueued before the `name` column existed (i.e. mid-upgrade) are reported under the name
+  `"unknown"`.
+- Because this metric's cardinality scales with the number of distinct job names, it can be
+  disabled if that's a concern for your metrics provider:
+
+```ruby
+Delayed::Monitor.emit_max_age_by_name = false
+```
+
+All of these events (including `max_age_by_name`) may be subscribed to via a single regular
+expression (again, in your application config or in an initializer):
 
 ```ruby
 ActiveSupport::Notifications.subscribe(/delayed\.job\..*_(count|age|percent)/) do |*args|
--- a/README.md
+++ b/README.md
@@ -417,10 +417,10 @@ QUEUES=mailers,tasks rake delayed:monitor
 ```
 
 The following events will be emitted, grouped by priority name (e.g. "interactive"), queue name,
-and the values of any configured `tag_columns`. By default, job `name` is included. The
-metric's "`:value`" will be available in the event's payload.  **This means that there will be one
-value _per_ unique combination of queue, priority, and tag column values**, and totals must be
-computed via downstream aggregation (e.g. as a StatsD "gauge" metric, summed or maxed by tag).
+and the values of any columns configured via `tag_columns` (none, by default). The metric's
+"`:value`" will be available in the event's payload.  **This means that there will be one value
+_per_ unique combination of queue, priority, and tag column values**, and totals must be computed
+via downstream aggregation (e.g. as a StatsD "gauge" metric, summed or maxed by tag).
 
 - **delayed.job.count** - the total number of jobs
 - **delayed.job.future_count** - jobs where run_at is in the future
@@ -436,15 +436,13 @@ An additional _experimental_ metric is available, intended for use with applicat
 
 - **delayed.job.alert_age_percent** - the _percent_ to which the oldest job has reached the "age alert" threshold. (See the [Alerting Threshholds](#priority-based-alerting-threshholds) section above.)
 
-By default, these events are also tagged with the job's `name` (when the jobs table has a `name`
-column — see [Database Setup](#database-setup)) so that downstream aggregation can answer
-"_which_ job is stuck?" when `delayed.job.max_age` alerts (e.g. `max by {queue, priority, name}`
-in Datadog).
+#### Tagging metrics with additional columns
 
-The set of tagged columns is driven by `Delayed::Monitor.tag_columns`, which defaults to
-`%i(name)` when the jobs table has a `name` column (and to `[]` otherwise). You can include
-columns your application adds to the jobs table (populated at enqueue time). For example, if your
-jobs table has an `owner` column you wish to also monitor:
+These events can also be tagged with the values of other jobs-table columns, so that downstream
+aggregation can answer "_which_ job is stuck?" when `delayed.job.max_age` alerts (e.g.
+`max by {queue, priority, name}` in Datadog). This is opt-in via `Delayed::Monitor.tag_columns`,
+which defaults to `[]`. To tag by the job's `name` (see [Database Setup](#database-setup)) plus an
+`owner` column your application has added to the jobs table and populates at enqueue time:
 
 ```ruby
 Delayed::Monitor.tag_columns = %i(name owner)
@@ -452,6 +450,10 @@ Delayed::Monitor.tag_columns = %i(name owner)
 
 A few behavioral notes:
 
+- Each tag column is added to the monitor's `GROUP BY`, and the generated indexes (e.g.
+  `idx_delayed_jobs_live`) cover only `priority` and `queue`. Grouping by anything else can push
+  these queries off their index and into significantly more expensive scans, so evaluate any tag
+  column by checking the query plans against a production-sized jobs table.
 - Rows whose value was never populated for a tagged column are reported under the value `'unset'`
   (e.g. jobs enqueued before the `name` column existed, mid-upgrade).
 - Configured columns must exist on the jobs table: assigning a missing column to `tag_columns`
@@ -461,8 +463,8 @@ A few behavioral notes:
 - Tag values cannot be enumerated in advance, so a tagged series is only emitted while matching
   jobs are present. Separately, an untagged zero value is always emitted for every
   (priority, queue) combination, so that each metric maintains a baseline series even when no
-  matching jobs are enqueued. For example, `delayed.job.count` with a single enqueued job would
-  emit the following series:
+  matching jobs are enqueued. For example, with `tag_columns = %i(name)` and a single enqueued
+  job, `delayed.job.count` would emit the following series:
 
   ```ruby
   { priority: 'interactive', queue: 'default', name: 'SimpleJob', value: 1 }
@@ -473,13 +475,12 @@ A few behavioral notes:
   ```
 - Each column multiplies a metric's series cardinality by its number of distinct values (though in
   practice a job's `name` tends to determine its `priority` and any ownership tags, making the
-  number of distinct job names the effective upper bound). If cardinality is a concern for your
-  metrics provider, tagging can be disabled entirely with `Delayed::Monitor.tag_columns = []`.
+  number of distinct job names the effective upper bound).
 
-#### Rolling out a new tag column
+#### Rolling out a tag column
 
-Because assignment fails loudly on a missing column, a new tag column should be rolled out in
-three separate deploys, each fully released before the next begins:
+Because assignment fails loudly on a missing column, a tag column should be rolled out in three
+separate deploys, each fully released before the next begins:
 
 1. Migrate the column onto the jobs table (nullable — no backfill required).
 2. Deploy the code that populates the column at enqueue time.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -423,7 +483,7 @@ ActiveSupport::Notifications.subscribe(/delayed\.job\..*_(count|age|percent)/) d
 end
 ```
 
-Additionally, the monitor process with emit a **delayed.monitor.run** event with a duration
+Additionally, the monitor process will emit a **delayed.monitor.run** event with a duration
```
BetterBoyer, line 486: This seemed like a typo so I just updated it

README.md
```diff
@@ -409,8 +411,66 @@ An additional _experimental_ metric is available, intended for use with applicat
 
 - **delayed.job.alert_age_percent** - the _percent_ to which the oldest job has reached the "age alert" threshold. (See the [Alerting Threshholds](#priority-based-alerting-threshholds) section above.)
 
-All of these events may be subscribed to via a single regular expression (again, in your application
-config or in an initializer):
```
BetterBoyer, line 413: Did not remove, git diff is just not recognizing it moved below the new block of comments added for the new feature

README.md
```diff
@@ -391,14 +391,16 @@ QUEUE=tracking rake delayed:monitor
...
+The following events will be emitted, grouped by priority name (e.g. "interactive"), queue name,
+and the values of any configured `tag_columns`. By default, job `name` is included. The
+metric's "`:value`" will be available in the event's payload.  **This means that there will be one
+value _per_ unique combination of queue, priority, and tag column values**, and totals must be
+computed via downstream aggregation (e.g. as a StatsD "gauge" metric, summed or maxed by tag).
 
 - **delayed.job.count** - the total number of jobs
 - **delayed.job.future_count** - jobs where run_at is in the future
 - **delayed.job.working_count** - jobs that are currently being worked off (excludes failed jobs)
+- **delayed.job.locked_count** - jobs that are currently locked by a worker (equivalent to working_count)
```
BetterBoyer, line 403: This was missing so added it in

joey-betterment (dismissed):
LGTM

I think it would be a good idea to make sure grouping by name won't cause any slowdowns. Might be an expensive query on a larger table. Otherwise, :shipit: !

tkoar (dismissed):
LGTM 🚀 

Thanks for tackling the custom columns inclusion 🙇

lib/delayed/monitor.rb
```diff
@@ -17,6 +17,19 @@ class Monitor
 
     cattr_accessor :sleep_delay, instance_writer: false, default: 60
 
+    def self.tag_columns
+      @tag_columns ||= (Job.column_names.include?('name') ? %i(name) : []).freeze
```
smudge: Per an earlier convo, I think we should remove `'name'` from the defaults (for now) and make that change in a future update where we also revise the way we construct the default indexes (e.g. `idx_delayed_jobs_live`). Without a way to ensure parity between query grouping and index ordering, we see regressions in both MySQL and PostgreSQL query plans for the two biggest queries (live counts + failed counts):

Image

Image

Image

My read at a glance is that we'd see more instances of current "index only" scans instead requiring table lookups (or in one instance on MySQL, a full table scan), and this will have an immediate impact on how well this monitor behaves on a very large or high-throughput jobs table.

smudge (changes requested):
Apologies for the slow review turnaround!

lib/delayed/monitor.rb
```diff
@@ -17,6 +17,19 @@ class Monitor
...
 
+    def self.tag_columns
+      @tag_columns ||= [].freeze
+    end
+
+    def self.tag_columns=(columns)
+      if columns.any? { |column| Job.column_names.exclude?(column.to_s) }
+        raise ArgumentError, "Delayed::Monitor.tag_columns includes columns missing from #{Job.table_name}. " \
+                             "Available columns: #{Job.column_names.join(', ')}"
+      end
```
smudge: I'm wondering if we should defer this check, so that `Delayed::Monitor.tag_columns=` can be called without needing an active/valid DB connection or loading of classes in `app/models`. (Specifically thinking about initializer/boot/config code.)

lib/delayed/monitor.rb
```diff
@@ -67,10 +80,13 @@ def self.parse_utc_time(string)
     attr_reader :jobs
 
     def emit_metric!(metric)
-      query_for(metric).reverse_merge(default_results).each do |(priority, queue), value|
+      query_for(metric)
+        .merge!(default_results) { |_key, existing, _default| existing }
+        .each do |(priority, queue, *column_values), value|
+        tags = column_values.zip(self.class.tag_columns).to_h { |val, column| [column, val.nil? ? 'unset' : val] }
```
smudge: We should be okay accepting `nil` here as the value:

```suggestion
        tags = column_values.zip(self.class.tag_columns).to_h { |val, column| [column, val] }
```

I'm thinking it would be best for consumers of the instrumentation to decide how to map `nil` tags. For example, when submitting to datadog, I think `"key:"` is a valid submission that would map to `""` or `"none"` depending on the metric type.

lib/delayed/monitor.rb
```diff
@@ -96,24 +112,30 @@ def default_tags
...
       Delayed::Job
-        .from(scope.select(:priority, :queue, *inner_selects).group(:priority, :queue))
-        .group(priority_case_statement, :queue).select(
+        .from(scope.select(:priority, :queue, *extra_group_columns, *inner_selects).group(:priority, :queue, *extra_group_columns))
+        .group(priority_case_statement, :queue, *extra_group_columns).select(
           *outer_selects,
           "#{priority_case_statement} AS priority",
           'queue AS queue',
-        ).group_by { |j| [j.priority.to_i, j.queue] }
+          *extra_group_columns.map { |column| "#{column} AS #{column}" },
```
smudge:
```suggestion
          *self.class.tag_columns,
```

It would be highly unlikely for these column names to be end-user-supplied, but in the off chance we haven't anticipated some use case for this config, passing them in as symbols avoids any SQL injection.

lib/delayed/monitor.rb
```diff
@@ -175,6 +197,7 @@ def oldest_workable_job_grouped
     def live_counts
       @memo[:live_counts] ||= grouped_query(
         jobs.live,
+        extra_group_columns: self.class.tag_columns,
```
smudge:
```suggestion
```

Can we reference `self.class.tag_columns` directly in the query construction, rather than passing them in?

README.md
```diff
@@ -409,8 +411,67 @@ An additional _experimental_ metric is available, intended for use with applicat
...
 
-All of these events may be subscribed to via a single regular expression (again, in your application
-config or in an initializer):
+#### Tagging metrics with additional columns
+
+These events can also be tagged with the values of other jobs-table columns, so that downstream
+aggregation can answer "_which_ job is stuck?" when `delayed.job.max_age` alerts (e.g.
+`max by {queue, priority, name}` in Datadog). This is opt-in via `Delayed::Monitor.tag_columns`,
+which defaults to `[]`. To tag by the job's `name` (see [Database Setup](#database-setup)) plus an
+`owner` column your application has added to the jobs table and populates at enqueue time:
```
smudge:
```suggestion
By default, the monitor only groups events by `priority` and `queue`. To add additional columns to the query's `GROUP BY` clause, declare them in an initializer config:
```

README.md
```diff
@@ -409,8 +411,67 @@ An additional _experimental_ metric is available, intended for use with applicat
...
+1. Migrate the column onto the jobs table (nullable — no backfill required).
+2. Deploy the code that populates the column at enqueue time.
+3. Add the column to `Delayed::Monitor.tag_columns` in an initializer, and deploy.
+
+Adding the column to `tag_columns` before the migration has run everywhere would raise at boot in
+every process that loads the initializer. Jobs enqueued before step 2 will report under the
+`'unset'` tag value until they are worked off (or backfilled).
+
+All of these events may be subscribed to via a single regular expression (again, in your
+application config or in an initializer):
```
smudge:
```suggestion
Avoid choosing columns with many unique values (e.g. `id`) as this will result in very poor query performance (and may essentially turn every row into its own metric result!). Instead, aim for lower-cardinality groupings such as `name` (the name of the job class) that are worth the query performance trade-off.

When adding `tag_columns`, **you are strongly encouraged to add new indexes** that incorporate these columns (rather than relying on `idx_delayed_jobs_live` and `idx_delayed_jobs_failed` for only partial coverage).

All of these events may be subscribed to via a single regular expression (again, in your application
config or in an initializer):
```

smudge (changes requested):
A few minor implementation notes. The biggest one is just the risk of `Delayed::Monitor.tag_columns=` needing access to the model & DB in order to be called, as it restricts where it can be used (e.g. if run in an initializer, would be nice for it to be cheap so it doesn't block or add live-DB requirements to some `rake` commands or deployment steps).

And I made some significant edits to the README, which I'm happy to discuss IRL 😅 -- mainly trying to give the feature a much smaller text footprint in the main README. If we want to we can add separate docs or wiki pages to drill further down into some of those behavioral notes, but it felt like "avoid high cardinality" and "consider your index coverage" were the two points we really needed to communicate about this feature.

lib/delayed/monitor.rb
BetterBoyer: That's fair. I was thinking initially that this was a case where we are capturing data with tagged names where we previously didn't have them and distinguishing that nil in the previous emission was not because it was nil but more because it was simply not getting recorded before. But this could have issues and I also agree that `'unset'` is making the decision for people which we probably don't want to do.

lib/delayed/monitor.rb
BetterBoyer: The initializer works I think for this since it should be booted with the Rails environment at that point. I was initially thinking it would be better to error at the moment you try to configure rather than later on. The only real cost I guess is not seeing the error at boot but later when the monitor is starting - but that tradeoff lands us in a more reliable spot anyway so fixing it

lib/delayed/monitor.rb
BetterBoyer: Yeah we can do that. I'm going to alter in one way though, just to handle the unlikely chance that someone tries to reassign `tag_columns` after initialization, I'm going to use a private attr instead that will be assigned in the initializer too

lib/delayed/monitor.rb
BetterBoyer: Addressed [here](https://github.com/Betterment/delayed/pull/117/changes/a816d2f2c746a92136230c3fd7f4c93c4c76365d)

lib/delayed/monitor.rb
BetterBoyer: Addressed [here](https://github.com/Betterment/delayed/pull/117/changes/3f07d41e74f6f409815be1de54dac75b7a48dfa3)

lib/delayed/monitor.rb
BetterBoyer: Addressed [here](https://github.com/Betterment/delayed/pull/117/commits#:~:text=making%20tag_columns%20instance%20variable) and [here](https://github.com/Betterment/delayed/pull/117/changes/7e1b9d7058ade36cc0635a2aa5c145896aae7fca)

lib/delayed/monitor.rb
BetterBoyer: Addressed [here](https://github.com/Betterment/delayed/pull/117/changes/7e1b9d7058ade36cc0635a2aa5c145896aae7fca)

README.md
```diff
@@ -458,6 +461,40 @@ ActiveSupport::Notifications.subscribe('delayed.monitor.run') do |*args|
...
+{ priority: 'eventual', queue: 'default', value: 0 }
+{ priority: 'reporting', queue: 'default', value: 0 }
+```
+
+Tag columns must already exist on the jobs table — the monitor validates this at startup and
+raises an `ArgumentError` if any are missing, so migrate a new column before adding it here.
+
+`NULL` values are emitted as `nil` tags. Your notification subscriber can decide how to represent
+these. Expect `nil`s for jobs enqueued before a newly added column was populated.
+
```
smudge, line 490:
```suggestion
```

This particular note feels self-evident, but I'm happy to leave it in if there's a pitfall you've identified that I'm not quite seeing.

smudge (approved):
LGTM, and thank you for this feature! 👏  Looking forward to trying it out.

Left one nitpick (take it or leave it).

PR #119, opened 2026-08-04
fix(ActiveJob): handle explicitly-nil priority

smudge:
It appears that newer versions of ActiveJob's retry_on always pass priority (nil unless specified) to retry_job, so `opts.key?(:priority)` coerced `nil` into `0`.

This means that every `retry_on` retry was silently re-enqueued at priority 0 (unless otherwise specified in the `retry_on` declaration).

### DIFF
--- a/spec/delayed/active_job_adapter_spec.rb
+++ b/spec/delayed/active_job_adapter_spec.rb
@@ -153,6 +153,12 @@ RSpec.describe Delayed::ActiveJobAdapter do
       expect(enqueued_delayed_jobs.last.priority).to eq(20)
     end
 
+    it 'ignores a nil priority, applying the default instead' do
+      JobClass.set(priority: nil).perform_later
+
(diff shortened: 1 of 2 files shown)


spec/delayed/active_job_adapter_spec.rb
```diff
@@ -153,6 +153,12 @@ def serialize(*)
       expect(enqueued_delayed_jobs.last.priority).to eq(20)
     end
 
+    it 'ignores a nil priority, applying the default instead' do
+      JobClass.set(priority: nil).perform_later
+
+      expect(enqueued_delayed_jobs.last.priority).to eq(10)
```
samandmoore, line 159: we could use a reference to the default priority here if we wanna make this tolerable of change to that value.

samandmoore (dismissed):

samandmoore (dismissed):

samandmoore (approved):

PR #122, opened 2026-08-04
feat(ActiveJob): preserve enqueued_at as created_at

smudge:
When a job is re-enqueued from previously-serialized job data (most notably a `retry_on`) the new row's created_at was stamped at insert time, so each retry reset the job's apparent age.

This change defaults a row's `created_at` to the wrapped job's (ActiveJob-maintained) enqueued_at, keeping the original creation time across the retry chain.

Notably, `created_at` no longer means "row insert time," so anything treating it as would now measure against the pre-retry enqueue time (which is ultimately the point of this change, but is a breaking change nonetheless).

We do **not** query `created_at` as part of the built-in `delayed:monitor` process, so this should not impact any standard gem installs.

### DIFF
--- a/lib/delayed/backend/job_preparer.rb
+++ b/lib/delayed/backend/job_preparer.rb
@@ -13,6 +13,7 @@ module Delayed
         set_queue_name
         set_priority
         set_run_at
+        set_created_at
         set_name
         handle_dst
         reject_stale_run_at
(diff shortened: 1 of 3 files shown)


samandmoore (dismissed):
noting the breaking change but agreed this doesn't give me any concerns

samandmoore (approved):

PR #127, opened 2026-08-06
feat(ActiveJob): Fail jobs when retry_on attempts are exhausted

smudge:
A `retry_on` declaration is an explicit retry budget for the errors it covers, so once its attempts are exhausted, further retries by the worker's own backstop are contrary to the developer's declared policy. Now, when a retry_on budget runs out, the job is immediately marked as failed.

The worker's retry backstop (`max_attempts`, 25 by default) still applies to errors that have no retry_on policy at all, and jobs may still retry for other reasons (including other retry_on declarations with remaining budget).

### DIFF
--- a/lib/delayed/job_wrapper.rb
+++ b/lib/delayed/job_wrapper.rb
@@ -61,7 +61,17 @@ module Delayed
       # The error escaped retry_on (if any), so ActiveJob considers the job terminated.
       job_data['terminated_at'] = record.class.db_time_now.utc.iso8601(9)
       record.payload_object = self # re-serialize the handler
+
+      # If the error escaped ActiveJob's retry_on policy, we stop the Delayed::Job retries as well.
+      @_aj_retry_terminated = rescue_handlers.any? { |name, _| error.is_a?(name.constantize) }
+
       super
+    end
+
+    def max_attempts
+      return 1 if @_aj_retry_terminated
+
+      super if respond_to_missing?(:max_attempts)
     end
 
     def perform
--- a/README.md
+++ b/README.md
@@ -232,6 +232,9 @@ together or independently:
   indicate a code bug or data issue that needs to be resolved before the job can
   succeed.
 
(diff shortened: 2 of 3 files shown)


lib/delayed/job_wrapper.rb
```diff
@@ -61,9 +61,19 @@ def error(record, error)
...
       record.payload_object = self # re-serialize the handler
+
+      # If the error escaped ActiveJob's retry_on policy, we stop the Delayed::Job retries as well.
+      @_aj_retry_terminated = rescue_handlers.any? { |name, _| error.is_a?(name.constantize) }
+
       super
     end
 
+    def max_attempts
+      return 1 if @_aj_retry_terminated
```
samandmoore, line 72: so is this temporally coupled to the error lifecycle hook? is that kosher in this case? it's usually a smell but it's hard to reason about whether it's okay in this case.

smudge: Yeah, it's a little clever, but it solves for an ergonomic pain point -- imagine you've set up a `retry_on` policy for `SomeTransientError` telling it to retry N times before giving up, only to find that once `retry_on` hits its own attempt counter, the `delayed` framework itself keeps retrying for its own default `max_attempts` of M, so now you have N*M retries. (Or maybe N + M, if ActiveJob never resets its internal retry counter... it gets hard to reason about.)

You could override `max_attempts` for that job to 1, but if a job fails due to something else, then without a `retry_on` policy to catch it you don't get any retries at all.

So this effectively says, if you have a `retry_on` policy defined for a specific combination of job & exception type, then `delayed` will not retry on its own beyond that policy. If you do _not_ have a `retry_on` policy for a given exception, then it's an unhandled exception and `delayed`'s default retry policy will kick in (as will its error reporting, so queue health goes down for unhandled exceptions, but does not go down if a `retry_on` policy is in place, as this retry is invisible to `delayed`.)

samandmoore (approved):

PR #132, opened 2026-08-12
feat: Ruby 4.0 and Rails 8.1 support

smudge:
This adds up to Ruby 4.0 and up to Rails 8.1 to the test matrix, and addresses #131 in the process.

(No changes were necessary to maintain EOL version support -- only a change from Rails 8.0 to 8.1 -- so I did not change any lower bounds or remove older versions from the matrix.)


.github/workflows/ci.yml
```diff
@@ -50,7 +63,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2
-    - uses: ruby/setup-ruby@a4effe49ee8ee5b8b5091268c473a4628afb5651 # v1
+    - uses: ruby/setup-ruby@6e5d382445ae5590b7449d8b3bc8cb1c2c27f617 # v1
```
smudge, line 66: Pinning [here](https://github.com/ruby/setup-ruby/commit/6e5d382445ae5590b7449d8b3bc8cb1c2c27f617) for Ruby 4.0.* availability.

spec/helper.rb
```diff
@@ -299,6 +299,8 @@ def formatted
       .gsub(/AS ("|`)?(\w+)("|`)?/) { "AS #{Regexp.last_match(2)[0...63]}" }
       # newline and indent when aliased columns are listed
       .gsub(/AS (\w+),/) { "AS #{Regexp.last_match(1)},\n      " }
+      # normalize 'FROM (...) AS' subqueries (changes across Rails versions)
+      .gsub(") AS subquery") { ") subquery" }
```
smudge, line 303: This variance appears to only affect the sqlite query construction, and changes from Rails 8.0 to 8.1.

(This is all just for snapshot testing, but at some point we'll just have to parse & dump the SQL rather than hand-normalizing via regex 😅)

effron (approved):

PR #133, opened 2026-08-13
fix: Defer Rails constant references so `require 'delayed'` works standalone

frewsxcv:
As of 4.0.0, a bare `require 'delayed'` raises `NameError` unless activerecord (and activejob) are already loaded, which bites tools that require gems in isolation (e.g. tapioca's gem-loading phase):

1. `lib/delayed/helpers/migration.rb` evaluates `RETRY_EXCEPTIONS = [ActiveRecord::LockWaitTimeout, ...]` in the module body, before `lib/delayed.rb` requires `active_record`. Now a lazily memoized `Migration.retry_exceptions` method (the constant was only used internally). Bonus: the `defined?(PG::LockNotAvailable)` check now runs at migration time, so it works even when `pg` loads after `delayed`.
2. `lib/delayed/job_wrapper.rb` calls `ActiveJob.version` at class definition time, but `lib/delayed.rb` never requires `active_job` (integration is deferred via `on_load(:active_job)`). The Rails 8.1 check now happens at rescue time via `defined?(ActiveJob::UnknownJobClassError)`, with unchanged semantics (`UnknownJobClassError` subclasses `NameError`).

Includes a regression spec that requires `delayed` in a subprocess with no frameworks pre-required (the existing suite can't catch this, since `spec/helper.rb` preloads all frameworks). No eager requires added, preserving the gem's lazy-loading design.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    delayed (4.0.0)
+    delayed (4.0.1)
       activerecord (>= 6.0)
       benchmark
       concurrent-ruby
--- a/lib/delayed/helpers/migration.rb
+++ b/lib/delayed/helpers/migration.rb
@@ -33,7 +33,6 @@ module Delayed
         end
       end
 
-      # Evaluated lazily so that requiring this file does not require active_record.
       def self.retry_exceptions
         @retry_exceptions ||= [
           ActiveRecord::LockWaitTimeout,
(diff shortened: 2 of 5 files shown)


smudge (dismissed):
Great catch, and thank you! (Last night I suddenly had the thought that we don't have any `Appraisal` for non-ActiveJob usage.) If you want to update `version.rb` to cut a patch release number, we can release as soon as this PR merges!

smudge (approved):

PR #134, opened 2026-08-19
chore: test against postgres 18 in CI

effron:
Bump the CI postgres service from 15 to 18, update the snaspshot to reference pg 18 behavior.


smudge (approved):
Nice to see that for multiple queues it now includes them in the index condition. 👍

